PR #23, opened 2019-01-30
perf improvements [wip]

mpenet:
small gains, just getting familiar with the codebase and replacing bit along the way: 

* most doseqs can be replaced with run! 
* lots of (zipmap xs (range)) can be more efficient via into + xform
* ...


mpenet:
I ll open that on datascript as it's shared code apparently.

whilo:
Yeah, we will rebase soonish. We first want to have a design for good schema support in datahike.

PR #29, opened 2019-03-07
28-datascript-rebase

kordano:


kordano:
fixes #28

src/datahike/db.cljc
```diff
@@ -521,6 +525,7 @@
                  create-avet (fn [a v e tx] (Datom. e a v tx true))
                  ]
              (case-tree [e a (some? v) tx]
+                        ;;rb: shorter datom syntax
```
whilo: We only need to pass vectors for datoms there. The Datom construction is unnecessary, I was exploring to store datoms in the index instead of flat vectors, but I think the latter is better (because serialization is more compact and easier that way).

whilo:
I think it would be good to drop the stale `:eavt`, `:aevt` and `:avet` index references to the in-memory DB at the end of the rebase and move all the `*-durable` ones into their place. This was only helpful for the initial porting. 

Ideally we would also move our hitchhiker-tree backend of the IDB protocol probably in a separate file and leave the original one with @tonsky's optimized btset in place, so that we can use it for our memory backend as well. This will require lifting of database creation and transaction functions (empty-db, init-db, with-datom, ...) to be polymorphic on the backend, since currently the code only dispatches on the hitchhiker-tree's backend implementation internally. This is not strictly necessary, but the hitchhiker-tree is significantly slower for in-memory database creations and a few times slower for lookups, so it would pay off in the long run. What do you think?

kordano:
You're right. I'll planned to do that after I fixed the current rebase tests.

kordano:
I would use the current as-is rebase and create a new issue for the refactoring and generalization.

project.clj
```diff
@@ -1,43 +1,38 @@
-(defproject io.replikativ/datahike "0.1.3"
+(defproject io.replikativ/datahike "0.2.0-beta"
   :description "A durable datalog implementation adaptable for distribution."
   :license {:name "Eclipse"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :url "https://github.com/replikativ/datahike"
+  :url "https://github.com/tonsky/datahike"
```
whilo: This needs a fix :).

kordano:
I'll remove `:eavt-durable` with #32

Issue #38, opened 2019-04-25
hash changes for the same transaction


kordano:
The same transaction on a db value results in different hash values: 

```clojure
(:require '[datahike.core :as d])

(def db (d/db-with (d/empty-db) [{:db/id 1 :name "Konrad"}]))
(hash (d/db-with db [[:db.fn/retractEntity 1]]))
;; => 32706373
(hash (d/db-with db [[:db.fn/retractEntity 1]]))
;; => 13589158
```

whilo:
The hitchhiker-tree has mutable helper objects attached to tree nodes, maybe its hashes break equality.

PR #39, opened 2019-05-03
Add failing test for db hash changes

purrgrammer:
Not sure if you wanted to address #38 as part of the index protocols PR, feel free to close if not.


kordano:
Many thanks @purrgrammer , I'll add it to the index protocols PR.

PR #48, opened 2019-06-25
Use hash of datoms instead of EAVT index for calculating DB hash

purrgrammer:


purrgrammer:
Should make #37 pass all the tests. /cc @kordano

kordano:
Thanks for your contribution @purrgrammer , I'll review it in the evening.

kordano:
Looks good, thanks again.

whilo:
This fix requires to iterate over all datoms to create the hash, this means it is O(n) in the size of the database. It will probably be used in many internal functions of Clojure or even the JVM, e.g. HashMaps. The right way is to fix the hash calculation of the EAVT index object.

whilo:
Ok, it is cached in this context, so probably it is fine. It is still curious that it is not behaving right for the EAVT index.

whilo:
The problem is that equality for `defrecord` in the hitchhiker-tree implementation tracks mutable values that differ and are accounted by the default `.equals` and `.hashCode` implementations.

whilo:
@purrgrammer @kordano The `O(n)` cost is definitely unnecessary and can make the whole system hang for arbitrary length of time in unintuitive situations where `.hashCode` is called internally. 

So the proper way to fix this, I think, is to calculate a rolling hash and store it within the root. Since the order of insertions does not matter it is enough to calculate the hash of each datom on insertion and add it to the current DB hash modulo `Long.MAX_VALUE` (or just accepting overflow). This will then work independent from index implementation. @purrgrammer Could you provide a patch for this?

purrgrammer:
> So the proper way to fix this, I think, is to calculate a rolling hash and store it within the root. Since the order of insertions does not matter it is enough to calculate the hash of each datom on insertion and add it to the current DB hash modulo Long.MAX_VALUE (or just accepting overflow). This will then work independent from index implementation.

I agree, sounds good!

>  @purrgrammer Could you provide a patch for this?

Sure.

PR #49, opened 2019-06-25
40 pull parser

kordano:


src/datahike/pull_api.cljc
```diff
@@ -1,12 +1,13 @@
...
-    [datahike.db :as db]
-    [datahike.datom :as dd]
-    [datahike.pull-parser :as dpp #?@(:cljs [:refer [PullSpec]])])
-    #?(:clj
-      (:import
-        [datahike.datom Datom]
-        [datahike.pull_parser PullSpec])))
+   [datahike.db :as db]
+   [datahike.datom :as dd]
+   #?@(:cljs [datalog.parse.type :refer [PullSpec]])
```
grischoun: Isn't this supposed to be `.parser.`?
And also `PullSpec` does not exist in `datalog.parser.type`.

src/datahike/pull_api.cljc
```diff
@@ -1,12 +1,13 @@
...
+   [datalog.parser.pull :as dpp])
```
grischoun: The alias `dpp` is never used later in the file.

src/datahike/pull_api.cljc
```diff
@@ -272,8 +273,8 @@
 
 (defn pull [db selector eid]
   {:pre [(db/db? db)]}
-  (pull-spec db (dpp/parse-pull selector) [eid] false))
+  (pull-spec db (datalog.parser.pull/parse-pull selector)  [eid] false))
```
grischoun: use `dpp` instead?

src/datahike/pull_api.cljc
```diff
@@ -272,8 +273,8 @@
...
 
 (defn pull-many [db selector eids]
   {:pre [(db/db? db)]}
-  (pull-spec db (dpp/parse-pull selector) eids true))
+  (pull-spec db (datalog.parser.pull/parse-pull selector) eids true))
```
grischoun: use `dpp` instead?

src/datahike/pull_api.cljc
grischoun: Also, is `type` the right name here?

kordano:
thanks for the feedback. I will fix the issues.

src/datahike/pull_api.cljc
kordano: You're correct, it is the wrong namespace.

src/datahike/pull_api.cljc
kordano: I adjusted that.

PR #51, opened 2019-06-27
Pluggable store through multimethods

purrgrammer:


kordano:
Looks good. I'll merge it to #55 as well and I'll adjust to the new `create-database` signatures.

PR #53, opened 2019-07-01
Compute rolling hash of datoms

purrgrammer:
- Hashes form a commutative group under addition
- A datom hash is computed using EAV so on retraction we substract the hash
- Remove hashing support from `FilteredDB`


kordano:
@whilo could you have a look at that?

whilo:
This looks good to me.

PR #58, opened 2019-07-26
Release 0.2.0

kordano:


test/datahike/test/schema.cljc
```diff
@@ -0,0 +1,300 @@
...
+      (is (= #{["Alice" :audi] ["Alice" :bmw] ["Bob" :chrysler] ["Bob" :daimler]}
+             (d/q '[:find ?o ?c :where [?e :owner ?o] [?e :cars ?c]] (d/db conn)))))
+
+    (testing "insert to cardinality one"
+      (d/transact! conn [{:db/id [:owner "Alice"]
+                          :owner "Charlie"}])
+      (is (= #{["Charlie" :audi] ["Charlie" :bmw] ["Bob" :chrysler] ["Bob" :daimler]}
+             (d/q '[:find ?o ?c :where [?e :owner ?o] [?e :cars ?c]] (d/db conn)))))
+
+    (testing "test  cardinality change if unique is set"
```
grischoun: There is an extra space before 'cardinality'.

test/datahike/test/schema.cljc
```diff
@@ -0,0 +1,300 @@
...
+
+
+(deftest test-schema-persistence
+  (testing "test file persistence"
+    (let [uri (str "datahike:file:///tmp/dh-test-persistence")
+          _ (d/delete-database uri)
+          _ (d/create-database {:uri uri :initial-tx [name-schema]})
+          conn (d/connect uri)]
+      (testing "schema exists on creation and first connection"
+        (is (= #{[:name :db.type/string :db.cardinality/one]} (d/q find-schema-q (d/db conn)))))
```
grischoun: `#{[:name :db.type/string :db.cardinality/one]}` could be extracted and reused.

test/datahike/test/schema.cljc
```diff
@@ -0,0 +1,300 @@
...
+      (testing "reconnect with db"
+        (let [new-conn (d/connect uri)]
+          (is (= #{[:name :db.type/string :db.cardinality/one]} (d/q find-schema-q (d/db new-conn))))))
+      (d/delete-database uri))))
+
+(deftest test-schema-on-read-db
+  (testing "test database creation wit schema-on-read"
```
grischoun: `wit`-> `with`

examples/schema-intro/src/example/core.clj
```diff
@@ -0,0 +1,98 @@
...
+
+;; change bobs name to bobby
+(d/transact! conn [{:db/id 8 :contributor/name "bobby"}])
+
+;; check it
+(d/q find-name-email (d/db conn))
+
+(d/pull (d/db conn) '[*] [:contributor/name "bobby"])
+
+;; bob is not related anymore as index
```
grischoun: I don't understand what this comment is trying to say.

src/datahike/index.cljc
```diff
@@ -0,0 +1,108 @@
...
+  (-count [index])
+  (-insert [index datom index-type])
+  (-remove [index datom index-type])
+  (-slice [index from to index-type])
+  (-flush [index backend])
+  (-transient [index])
+  (-persistent! [index]))
+
+
+(extend-type DataNode
```
grischoun: This is not related to an issue with the code but more something that I would like to understand. What is the difference between `DataNode`and `IndexNode`?

src/datahike/core.cljc
```diff
@@ -605,6 +593,21 @@
...
 
 
+;; Data Readers
+
+(def ^{:doc "Data readers for EDN readers. In CLJS they’re registered automatically. In CLJ, if `data_readers.clj` do not work, you can always do
+
+             ```
+             (clojure.edn/read-string {:readers data-readers} \"...\")
+             ```"}
+  data-readers {'datahike/Datom dd/datom-from-reader
```
grischoun: Isn't this a duplicate of `src/data-readers.clj`?

src/datahike/connector.cljc
```diff
@@ -0,0 +1,129 @@
...
+          "Creates a new database"
+          {:arglists '([config])}
+          (fn [config] (type config)))
+
+(defmethod create-database String [uri]
+  (create-database {:uri uri}))
+
+(defmethod create-database clojure.lang.PersistentArrayMap
+  [{:keys [uri initial-tx schema-on-read]}]
+  (let [[m store-scheme path] (parse-uri uri)
```
grischoun: Shouldn't `m` be `s`as we are referring to a 'scheme'?

src/datahike/connector.cljc
```diff
@@ -0,0 +1,129 @@
...
+        {:keys [eavt aevt avet schema rschema config]} (db/empty-db {:db/ident {:db/unique :db.unique/identity}} (ds/scheme->index store-scheme) :config config)
+        backend (kons/->KonserveBackend store)]
+    (<?? S (k/assoc-in store [:db]
+                       {:schema   schema
+                        :rschema  rschema
+                        :config   config
+                        :eavt-key (di/-flush eavt backend)
+                        :aevt-key (di/-flush aevt backend)
+                        :avet-key (di/-flush avet backend)}))
+    (ds/release-store store-scheme store)
```
grischoun: I find it a bit surprising that just after creating the store, we have to already release it. May be it is just a naming issue.

src/data_readers.clj
```diff
@@ -1,2 +1,3 @@
-{datahike/Datom datahike.db/datom-from-reader
+{datahike/Datom datahike.datom/datom-from-reader
```
grischoun: What is this map for actually?

src/datahike/db.cljc
```diff
@@ -1058,24 +600,38 @@
...
+(defn- validate-val [v [_ _ a _ _ :as at] db]
   (when (nil? v)
     (raise "Cannot store nil as a value at " at
-           {:error :transact/syntax, :value v, :context at})))
+           {:error :transact/syntax, :value v, :context at}))
+  (when-not (get-in db [:config :schema-on-read])
+    (let [schema (:schema db)
+          schema-spec (if (ds/schema-attr? a) ds/implicit-schema-spec schema)]
+      (when-not (ds/value-valid? at schema)
+        (raise "Bad entity value " v " at " at ", value does not match schema definition. Must be conform to: " (ds/describe-type (get-in schema-spec [a :db/valueType]))
```
grischoun: There is duplication between the error message creation here and in the test. May be the string creation should be moved into a function?

src/datahike/index.cljc
whilo: In B+-trees IndexNodes only manage branching and all the data is at leaves (DataNodes).

grischoun:
I see. Thanks.

examples/schema-intro/src/example/core.clj
purrgrammer: Since the name of `"bob"` has been changed to `"bobby'` the identity `[:contributor/name "bob"]` will not match anymore. Maybe the phrasing can be improved but that's the idea.

kordano: I'll re-phrase it in a better way.

src/datahike/schema.cljc
```diff
@@ -0,0 +1,136 @@
...
+         (case attr-def
+           :db/cardinality (if (= new-value :db.cardinality/many)
+                             (if (get-in attr-schema [:db/unique])
+                               (assoc m attr-def [old-value new-value])
+                               nil)
+                             (assoc m attr-def [old-value new-value]))
+           :db/unique (when-not (get-in attr-schema [:db/unique])
+                        (when-not (= (get-in attr-schema [:db/cardinality]) :db.cardinality/one)
+                          (assoc m attr-def [old-value new-value])))
+           (assoc m attr-def [old-value new-value]))))) {} (dissoc entity :db/id)))
```
whilo: I would line break here in front of `{}` to make the arguments of `reduce-kv` more easily visible.

src/datahike/schema.cljc
```diff
@@ -0,0 +1,136 @@
...
+(defn schema-entity? [entity]
+  (some #(contains? entity %) schema-keys))
+
+(defn schema? [schema]
+  (s/valid? ::schema schema))
+
+(defn describe-type [schema-type]
+  (s/describe schema-type))
+
+(defn find-invalid-schema-updates [entity attr-schema]
```
whilo: Why are these proper rules for schema upgrades? I assume it is following some datomic rational. It is important that downstream queries do not break after schema changes and it can be the case with datalog that allowing multiple values for an attribute makes a query return less (e.g. nothing if it checks for uniqueness itself). We do not strictly have to have a good schema-update strategy in place, because it is actually context-specific on how the database is read, which cannot be known in general, but this is something we at least need to document well.

src/datahike/connector.cljc
```diff
@@ -0,0 +1,159 @@
...
+(defn transact [connection tx-data]
+  (try
+    (deref (transact! connection tx-data))
+    (catch Exception e
+      (throw (.getCause e)))))
+
+(defmulti create-database
+  "Creates a new database"
+  {:arglists '([config & opts])}
+  (fn [config & opts] (type config)))
```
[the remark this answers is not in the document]
kordano: Fixed with #70

src/datahike/schema.cljc
kordano: Fixed with #70

src/datahike/schema.cljc
kordano: I will create some documents about database migrations in general.

kordano: Fixed with #70.

Issue #59, opened 2019-08-04
entity/touch entity/get not resolving db/ident


boxxxie:
I have a DB where i have a bunch of idents, and when i get an entity out of the DB and use `(get)` or `(touch)` which uses an ident ref, it will come back as a set of `:db/id`

Is the expected behaviour for get or touch in this case to give me back the translated value of the ident (in this case keywords)?

also, possibly related, when i use `(walk/postwalk identity entity)` this throws an error. this is something that works in datomic.

anyway, if idents are supposed to resolve the way i described, i can try to add that functionality, otherwise working with idents seems to be a big PITA.

boxxxie:
After looking up this topic for a while, i think this is the correct behaviour in terms of returning `{:db/ident :ident-name}` when being touched. to me it seems like it lowers the utility of idents.

however, the `(walk identity entity)` code throwing errors seems wrong. i've looked over the entity namespace, and i'm not really sure how to go about fixing this behaviour

kordano:
I'll have a look at it @boxxxie . Could you add some code example, so I can reproduce this behaviour? Thanks.

boxxxie:
I'll try up something soon that uses datomic and datahike side by side for this issue.

kordano:
Thanks.

PR #63, opened 2019-08-12
Support walking through entities

purrgrammer:
From #59 

> also, possibly related, when i use (walk/postwalk identity entity) this throws an error. this is something that works in datomic.

I'm not sure why `empty` was not supported in entities, but is the only thing needed to enable the walking functions in the `clojure.walk` namespace IIUC.


kordano:
perfect, I'll have a look.

TimoKramer:
Closing in favor of #276, did a rebase there. Sorry that it took so long @purrgrammer and thanks.

purrgrammer:
No problem @TimoKramer, I'm glad it got merged! Thanks for taking the time to rebase it.

PR #68, opened 2019-08-19
66 add missing datomic signatures

kordano:


src/datahike/query.cljc
```diff
@@ -380,56 +409,56 @@
...
-                        (reduce (fn [acc tuple2]
-                                  (let [key (key-fn2 tuple2)]
-                                    (if-some [tuples1 (get hash key)]
-                                      (reduce (fn [acc tuple1]
-                                                (conj! acc (join-tuples tuple1 keep-idxs1 tuple2 keep-idxs2)))
-                                              acc tuples1)
-                                      acc)))
-                          (transient []) tuples2)
-                        (persistent!))]
+  (let [tuples1 (:tuples rel1)
```
grischoun: better indent the let?

test/datahike/test.cljc
```diff
@@ -29,7 +30,7 @@
     datahike.test.schema
     datahike.test.serialization
     datahike.test.store
-    datahike.test.temporal-index
+    datahike.test.time_variance
```
grischoun: Should use a - instead of _ in `time_variance?

test/datahike/test/purge.cljc
```diff
@@ -0,0 +1,126 @@
...
+
+(deftest test-purge
+  (let [conn (create-test-db "datahike:mem://test-purge")]
+    (testing "retract datom, data is removed from current db and found in history"
+      (let [name "Alice"]
+        (d/transact conn [[:db/retract [:name "Alice"] :age 25]])
+        (are [x y] (= x y)
+                   true (nil? (find-age @conn name))
+                   25 (find-age (d/history @conn) name))))
+    (testing "purge datom from current index"
```
grischoun: Add ' and from history'

examples/basic/src/examples/time_variance.clj
```diff
@@ -0,0 +1,142 @@
+(ns examples.time-variance
```
grischoun: Isn't something like 'history' a better name for the namespace?

test/datahike/test/time_variance.cljc
```diff
@@ -1,9 +1,8 @@
-(ns datahike.test.temporal-index
+(ns datahike.test.time_variance
```
grischoun: Should be -  instead of _

examples/basic/src/examples/time_variance.clj
grischoun: Or may be 'time-travel'.

PR #71, opened 2019-09-09
Use core predicates in specs whenever possible

purrgrammer:


src/datahike/schema.cljc
```diff
@@ -5,8 +5,8 @@
 (s/def :db.type/id #(or (= (class %) java.lang.Long) string?))
 
 ;; db types
-(s/def :db.type/bigdec #(= (class %) java.math.BigDecimal))
-(s/def :db.type/bigint #(= (class %) java.math.BigInteger))
+(s/def :db.type/bigdec decimal?)
+(s/def :db.type/bigint integer?)
```
purrgrammer: Note that this makes possible to transact any integer as a `:db.type/bigint`, which is convenient but figured out I should point it out.

kordano: Thanks, I'll merge it and release it with 0.2.1 and try to improve that with the next specs I'll add through the `datahike.api` and `datahike.core` namespace.

PR #84, opened 2019-10-26
Added a spec for db.type/number

pdenno:
As discussed on the #datahike slack channel, sometimes you expect number data but aren't sure whether it is going to be float or int. For that I added a spec db.type/number and added it to the spec db.type/value spec.

### DIFF
--- a/src/datahike/schema.cljc
+++ b/src/datahike/schema.cljc
@@ -10,6 +10,7 @@
 (s/def :db.type/boolean boolean?)
 (s/def :db.type/double double?)
 (s/def :db.type/float float?)
+(s/def :db.type/number number?)
 (s/def :db.type/instant #(= (class %) java.util.Date))
 (s/def :db.type/keyword keyword?)
 (s/def :db.type/long #(= (class %) java.lang.Long))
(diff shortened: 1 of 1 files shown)


whilo:
Thank you, that makes sense. @kordano what do you think?

pdenno:
I just noticed last night that I'm still getting messages like this:

Bad entity value :db.type/number at [:db/add 23 :db/valueType :db.type/number], value does not match schema definition. Must be conform to: #{:db.type/instant :db.type/boolean :db.type/uuid :db.type/value :db.type/string :db.type/keyword :db.type/ref :db.type/bigdec :db.type/float :db.type/bigint :db.type/double :db.type/long :db.type/symbol} 

So maybe there is more to this than adding those two lines of code. I verified that datahike is at least storing both double and long when declared db.type/number. Maybe there's something on my side, because it seems to be working for a few days. I'll look into this more tonight and get back to you.

kordano:
I would merge it into `development`since we try to keep the master update to date with latest releases and development with snapshot versions.

pdenno:
Sorry I missed that note about the development branch. Thanks for datahike! It is quite nice.

PR #86, opened 2019-11-08
DynamoDB support.

csm:
Uses [konserve-ddb](https://github.com/csm/konserve-ddb).

* src/datahike/config.cljc (::backend): add :ddb.
  (do-uri->config): new multimethod, dispatch on backend.
  Provide :ddb implementation.
  (uri->config): defer to do-uri->config.
* src/datahike/store.cljc: add implementations for :ddb.
* project.clj: add konserve-ddb (as provided dependency).

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,7 +12,8 @@
                  [io.replikativ/hitchhiker-tree "0.1.4"]
                  [io.replikativ/superv.async "0.2.9"]
                  [io.replikativ/konserve-leveldb "0.1.2"]
-                 [io.lambdaforge/datalog-parser "0.1.0"]]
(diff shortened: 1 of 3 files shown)


csm:
#89 may be more interesting than this.

whilo:
In some sense you have parametrized `konserve` there so that it is able to cover the functionality in this branch as well, right?

csm:
Yes, with `konserve-ddb-s3` just pass something that returns true for all keys to `consistent-key` and everything would be stored in DynamoDB. Though it would be odd to have to pass in a valid S3 bucket even though it's never used.

kordano:
Since Datahike now uses external backend libraries over protocols I'll have to close this PR. I'll try to move it to a separate project.

PR #89, opened 2019-11-14
DynamoDB+S3 support

csm:
Via [konserve-ddb-s3](https://github.com/csm/konserve-ddb-s3), the `:db` key is stored in DynamoDB, and is atomically updated. All other keys are stored in S3.

This doesn't include some other changes I was trying out in datahike, namely always using an index node for the root of the tree, out of concerns for item size in DynamoDB. It might be necessary to use an index node for the root if the data node size is increased -- fewer, larger objects is generally better for S3 -- since there's a good chance that a large root data node will overflow the 400KB limit for items in DynamoDB.

* src/datahike/config.cljc (`::backend`): add `:ddb+s3`.
  (`::region`, `::table`, `::bucket`, `::database`): new specs.
  (`:datahike/config`): add `::region`, `::table`, `::bucket`, `::database`.
  (`uri->config*`): new multimethod, dispatch on `:backend`. Add default implementation.
  (`uri->config`): defer to `uri->config*`.
* src/datahike/ddb_s3_store.clj: new namespace.
* project.clj: add konserve-ddb-s3 dependency.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -12,14 +12,15 @@
                  [io.replikativ/hitchhiker-tree "0.1.5-SNAPSHOT"]
                  [io.replikativ/superv.async "0.2.9"]
                  [io.replikativ/konserve-leveldb "0.1.2" :scope "provided"]
-                 [io.lambdaforge/datalog-parser "0.1.0"]]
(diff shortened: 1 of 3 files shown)


kordano:
See #86

PR #91, opened 2019-11-14
Add a Java binding

grischoun:


java/src/datahike/java/Datahike.java
```diff
@@ -0,0 +1,153 @@
...
+
+    public static Long tempId(Keyword k, Long i) {
+        return (Long)tempIdFn.invoke(k, i);
+    }
+
+    public static IEntity entity(Object dConn, int eid) {
+        return (IEntity)entityFn.invoke(dConn, eid);
+    }
+
+    // TODO: this would also benefit from an Entity interface
```
whilo: What would the Entity interface work here?

grischoun: This is how the method is redefined using the IEntity interface:

public static Object entityDb(IEntity entity) {

grischoun:
> Overall it looks good and I would leave it to the judgement of first explorative Java users. I am not sure about all the interface details. Before we commit to them I would give it a more thorough look again. How exhaustive are the tests?

Just now, I have added asserts that were missing in the current tests. (They are available in this PR.)
True, there are some methods that still need to be tested. I'll add them soon.
I have also added the method 'dbWith as the currently available 'with method is not fully usable (it is missing the declaration and implementation of an ITxReport to be really useful).

whilo:
Nice work! I think it is important to track which tests are missing still, so that this information does not get lost. In a sense this is important high-level information about the constraints the implementation is supposed to have and helps other contributors. We should keep track of this in general.

whilo:
Is this ready to merge?

grischoun:
No. Let me finish adding the documentation and it will be ready to go.

grischoun:
> I would merge this right now, but I am not sure how to run the JUnit tests and I think they should be automatically run by leiningen with `lein test` such that the Java API is always tested to be consistent with core API changes.

Good point. I'll see how to add that.

grischoun:
It now runs the JUnit tests automatically when 'lein test' is run.

Issue #96, opened 2019-11-19
Overhaul ID generation


kordano:
Since version 0.2.0 datahike adds also meta-entities to the index with transaction IDS as entity IDs in order to bind these meta entities to all transaction entities. We need better index ranges and proper definitions now.

PR #100, opened 2019-11-27
Reintroduce updated migrate namespace and test

jsmassa:
Updated and added old migrate namespace

### DIFF
--- a/src/datahike/migrate.clj
+++ b/src/datahike/migrate.clj
@@ -1,0 +1,24 @@
+(ns datahike.migrate
+  (:require [datahike.api :as api]
+            [datahike.db :as db]
+            [clojure.java.io :as io]))
+
+(defn export-db
+  "Export the database in a flat-file of datoms at path."
+  [db path]
+  (with-open [f (io/output-stream path)
+              w (io/writer f)]
+    (binding [*out* w]
+      (doseq [d (datahike.db/-datoms db :eavt [])]
+        (when (not= (:a d) :db/txInstant)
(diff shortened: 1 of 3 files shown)


markusalbertgraf:
Hi,
I put :tx/user data on many transactions to check who changed what record when.
I also use :tx/comment for e.g. schema upgrades.
`(not= (:a d) :db/txInstant)` leaves me with only half of that information.
Is it best practice to discard that on migration?
There are certainly scenarios where that makes sense though I would like to keep that if I just update the db-software or implementation.
If it discards :db/txInstant, does it make sense to keep the rest of the transaction-entities?
Will the db run ok with partial entities imported and if :db/id 536870913 is free but some above are taken?
Or could this perhaps be optional?

Thank you for your work!

whilo:
This migration code here is to keep compatibility with the code from `0.1.*`, where we just exported the EAVT index. It is true that we should export all transaction data from now onwards, @kordano is working on that separately. Neither this nor tx data was properly recorded in ´0.1.*´. @kordano Can you open and track an issue about this?

kordano:
I'll create an issue on that. It is also related to #96

markusalbertgraf:
Thank you!!!

efraimmgon:
I realize I am late to the party, but why do `export-db` and `import-db` write and read from a file, instead of just returning and accepting a `coll` of `datoms`?

Wouldn't this make them more flexible?

kordano:
We could add that as an option.

PR #110, opened 2020-01-08
91 datomic migration

kordano:


whilo:
What does the removal of the timestamp cases imply?

whilo:
It also makes sense to not use the name `migrate`, but rather `migrate-from-datomic`. We already have the `migrate` namespace and I think it would fit in there.

PR #115, opened 2020-01-27
adds cas support for nil, i.e. inserting first value

### LINKED ISSUE #114: cas for nil should work
[Datomic docs](https://docs.datomic.com/cloud/transactions/transaction-functions.html#db-cas) state that a CAS works for initial value of nil. This allows to insert the first value transactionally. But today in datahike this fails

```clojure
(require '[datahike.api :as db])
(db/create-database "datahike:file:///tmp/dh")
(def c (db/connect "datahike:file:///tmp/dh"))
(db/transact c [{:db/ident :a :db/cardinality :db.cardinality/one :db/valueType :db.type/keyword :db/unique :db.unique/identity}
                          {:db/ident :b :db/cardinality :db.cardinality/one :db/valueType :db.type/keyword}])
(db/transact c [{:a :a}])
(db/transact c [[:db/cas 3 :b nil :b]])
```
```
1. Unhandled clojure.lang.ExceptionInfo
   Bad entity value :b at [:db/cas 3 :b nil :b], value does not match schema
   definition. Must be conform to: keyword?
```

Same happens for a db ref. Both blow up in transact-tx-data, one calling entid-strict and one validate-val. Before validating the data we'd need to check if the old value is nil and treat it as an insert. I looked at the [relevant source](https://github.com/replikativ/datahike/blob/a181cc6c97beed408d0a415b72cfc888460dd076/src/datahike/db.cljc#L1380) and have a basic idea of what needs to be done but am unfamiliar with the internals to whip it up quickly. If you could either implement this yourself or provide a bit of guidance what functions need to be called where I'd be very glad.

xificurC:
fixes #114 , or so I'd wish.

This is a sketch of what might be a solution. It needs a bit of a refactor, but before that I'd need to know how to create a proper test case because the one I added seems to be running on schema-on-read mode and actually passes before any changes to db.cljc are made

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -1382,20 +1382,25 @@
             (let [[_ e a ov nv] entity
                   e (entid-strict db e)
                   _ (validate-attr a entity db)
-                  ov (if (ref? db a) (entid-strict db ov) ov)
                   nv (if (ref? db a) (entid-strict db nv) nv)
-                  _ (validate-val nv entity db)
                   datoms (-search db [e a])]
-              (if (multival? db a)
-                (if (some (fn [^Datom d] (= (.-v d) ov)) datoms)
+              (if (nil? ov)
+                (if (empty? datoms)
                   (recur (transact-add report [:db/add e a nv]) entities)
-                  (raise ":db.fn/cas failed on datom [" e " " a " " (map :v datoms) "], expected " ov
-                         {:error :transact/cas, :old datoms, :expected ov, :new nv}))
-                (let [v (:v (first datoms))]
-                  (if (= v ov)
-                    (recur (transact-add report [:db/add e a nv]) entities)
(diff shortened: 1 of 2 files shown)


whilo:
I have not used this Datomic feature before, thanks for tackling this issue.

xificurC:
Thanks. Just to make sure - you haven't used the feature because you didn't have such needs or you used a different feature to achieve the same result?

whilo:
I just have not needed it in this form yet, but I think supporting it makes sense.

kordano:
I'll add this one also to the 0.2.2 release.

xificurC:
Hey @kordano , thanks for merging so quickly. Beware that as I noted the test I added was passing even without the changes applied in this PR. I suspected the tests are running in schema-on-read mode but I'm not sure. So the test needs fixing!

Issue #82, opened 2019-10-21
create record for config


kordano:

TimoKramer:
1. read config from file
  - storage connection
  - schema
1. merge config with envvars
1. validate config
1. save config to record

TimoKramer:
priorities of config:
1. defaults
1. file
1. env
1. connect arguments
file, env and connect arguments are optional and via defaults there will be running an in-memory store

TimoKramer:
```
{:store {}
 :schema {}}
```
will be top level configs for now, like the corresponding namespaces

TimoKramer:
how to do it in cljs?

whilo:
Hey Timo! Thanks for working on this. Ideally the data structure for the config should look the same in cljs. What do you mean exactly?

TimoKramer:
I think it should have the same structure...I was referencing my problem with knowing nothing about how to handle any kind of env in js/node. I switched to using https://github.com/weavejester/environ for now, because it hopefully just works there. It seems to consider the usage in clj and cljs at least.

PR #128, opened 2020-03-18
#82 create record for config

TimoKramer:
hi there,
maybe you can check my solution on the config issue? It is now thought that one has to load the config-namespace to create a config. I did that to reuse the connector-code and not break the old ways to handle configuration.
I also updated the config-docs...
best regards, Timo

P.S: I would squash first before you merge so please tell me if you are happy with this solution or what would be better.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -8,6 +8,7 @@
                  [org.clojure/clojurescript "1.10.597" :scope "provided"]
                  [persistent-sorted-set     "0.1.2"]
                  [org.clojure/tools.reader "1.3.2"]
+                 [environ "1.1.0"]
                  [io.replikativ/hitchhiker-tree "0.1.5"]
                  [io.replikativ/superv.async "0.2.9"]
                  [io.lambdaforge/datalog-parser "0.1.1"]]
--- a/src/datahike/config.cljc
+++ b/src/datahike/config.cljc
@@ -1,26 +1,107 @@
 (ns datahike.config
-  (:require [clojure.spec.alpha :as s])
+  (:require [clojure.edn :as edn]
+            [clojure.spec.alpha :as s]
+            [environ.core :refer [env]])
   (:import [java.net URI]))
 
-
-(s/def ::username string?)
-(s/def ::password string?)
-(s/def ::path string?)
-(s/def ::backend keyword?)
-(s/def ::host string?)
-(s/def ::port int?)
-(s/def ::uri string?)
+(s/def ::backend #{:mem :file :level :pg})
+(s/def ::username (s/nilable string?))
+(s/def ::password (s/nilable string?))
+(s/def ::path (s/nilable string?))
+(s/def ::host (s/nilable string?))
+(s/def ::port (s/nilable int?))
 
 (s/def ::schema-on-read boolean?)
 (s/def ::temporal-index boolean?)
 
-(s/def :datahike/config (s/keys :req-un [::backend]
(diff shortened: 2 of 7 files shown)


TimoKramer:
I improved the solution so that you don't have to load config-namespace and run load-config to use create/connect/delete/exist?-functions. Now the config var is queryable and config is loaded on start. with a reload-config function you can now update the config with args otherwise it is using the environ/systemprops.

project.clj
```diff
@@ -8,11 +8,13 @@
...
                  [persistent-sorted-set     "0.1.2"]
                  [org.clojure/tools.reader "1.3.2"]
+                 [environ "1.1.0"]
                  [io.replikativ/hitchhiker-tree "0.1.5"]
                  [io.replikativ/superv.async "0.2.9"]
                  [io.lambdaforge/datalog-parser "0.1.1"]]
 
-  :plugins [[lein-cljsbuild "1.1.7"]]
+  :plugins [[lein-cljsbuild "1.1.7"]
+            [lein-environ "1.1.0"]]
```
TimoKramer: das mache ich wieder raus

src/datahike/config.cljc
```diff
@@ -1,26 +1,89 @@
...
-(s/def ::host string?)
-(s/def ::port int?)
-(s/def ::uri string?)
+(s/def ::backend #{:mem :file :level :pg})
+(s/def ::username (s/nilable string?))
+(s/def ::password (s/nilable string?))
+(s/def ::path (s/nilable string?))
+(s/def ::host (s/nilable string?))
+(s/def ::port (s/nilable int?))
+(s/def ::uri (s/nilable string?))
```
TimoKramer: that's to be removed

src/datahike/config.cljc
```diff
@@ -1,26 +1,89 @@
...
 
 (s/def ::schema-on-read boolean?)
 (s/def ::temporal-index boolean?)
 
-(s/def :datahike/config (s/keys :req-un [::backend]
-                                :opt-un [::host ::port ::path ::username ::password]))
+(s/def :datahike/store (s/keys :req-un [::backend]
+                               :opt-un [::username ::password ::path ::host ::port]))
+(s/def :datahike/schema (s/keys :opt-un [::schema-on-read ::temporal-index]))
```
kordano, line 20: namespaces will be changed in the future. So we would have something like `:index/schema-flexibility` with better naming.

TimoKramer: should I make schema-on-read and temporal-index first level options? Means taking schema-record out.

TimoKramer:
FMPOV this PR is good to go.

whilo:
This looks in general good to me, but I am not an expert in configuration management. The main thing that I see missing (besides the missing SSL encryption of issue #133) are tests that check that when the configuration is done wrong a proper error is raised.

TimoKramer:
You're right @whilo. I will insert logging-statements commented out and implement logging as next issue. I will try to add postgresql-ssl-support as well but as a separate issue.

TimoKramer:
squashed and ready to merge

TimoKramer:
one more thing I was thinking about now regarding the postgresql-config:
Wouldn't it be better to have `:path` for `:file` and `:level` and a `:database`-key for postgresql? I mean having path for a postgresql-database is kind of hackish.

kordano:
You are right about these specific backend keys. I would solve that with a separate PR.

PR #135, opened 2020-04-16
#133 postgres connection ssl

### LINKED ISSUE #133: Postgres connections requiring SSL not supported
When a postgres connection url specifies options such as `:ssl true`, they are not passed on to jdbc driver. See https://github.com/clojure/java.jdbc#example-usage which specifies `:ssl` and `:sslfactory`.

This is critical if one wants to use a hosted Postgres Database like from [heroku](https://www.heroku.com/postgres), which requires ssl for connecting.

TimoKramer:
### Datahike now supports ssl-connections

Since PostgreSQL supports encrypted connections there is the
need to add this option to datahike as well. I added the
configuration and adapted the datahike-postgres
to the new options and the new configuration format.

seemed to make more sense to switch datahike-postgres
pass the configuration as map because konserve-pg is able
handle a map anyways. Converting a string to map to string
not seem reasonable so I switched the lib to use map as
without breaking the api.

- PR #6 in replikative/datahike-postgres
- Closes #133

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -22,14 +22,17 @@ The sources are resolved in following order:
 That means passing a config as argument overwrites java system properties and using java system properties overwrite environment variables etc. Currently the configuration map looks like this per default:
 
 ```
-{:store {:backend  :mem     ;keyword
-         :username nil      ;string
-         :password nil      ;string
-         :path     nil      ;string
-         :host     nil      ;string
-         :port     nil}     ;int
- :schema-on-read   false    ;boolean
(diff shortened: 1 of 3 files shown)


TimoKramer:
BTW https://gist.github.com/mrw34/c97bb03ea1054afb551886ffc8b63c3b is quite useful for testing locally

TimoKramer:
just rebased

TimoKramer:
not relevant anymore

PR #136, opened 2020-04-17
129 no history support

kordano:


src/datahike/db.cljc
```diff
@@ -876,6 +877,10 @@
           :cljs [^boolean indexing?]) [db attr]
   (is-attr? db attr :db/index))
 
+(defn #?@(:clj [^Boolean no-history?]
+          :cljs [^boolean indexing]) [db attr]
```
TimoKramer, line 881: why does this have a different name for cljs? wouldn't that mean that with-datom (line 1081) is failing in cljs?

TimoKramer: and it would look better when lined up like the functions above

kordano: I think, I made an error when copying something. Thanks for pointing that out @TimoKramer . I'll fix it and you could have another look.

PR #137, opened 2020-04-20
Add java interop in queries

vlaaad:


src/datahike/query.cljc
```diff
@@ -539,11 +542,44 @@
...
+                                                            (.getParameterTypes method)
+                                                            args-classes)))
+                                method))))
+                 (throw (ex-info (str (.getName this-class) "."
+                                      method-name "("
+                                      (str/join "," (map #(.getName ^Class %) args-classes))
+                                      ") not found")
+                                 {:this-class this-class
+                                  :method-name method-name
+                                  :args-classes args-classes})))))))
```
vlaaad, line 565: Note: I memoize the results of the method search to improve performance. 
Note: I use `.getMethods` with manual parameter arg checking instead of `.getMethod` because the latter returns result for exact parameter argument match instead of assignable match. For example, `(.getMethod String "contains" (into-array Class [String]))` can't find contains method because it expects `CharSequence` and not a `String` as input argument even though strings are char sequences.

src/datahike/query.cljc
```diff
@@ -539,11 +542,44 @@
...
+
+
+(defn- resolve-method [method-sym]
+  #?(:cljs nil
+     :clj (let [method-str (name method-sym)]
+            (when (= \. (.charAt method-str 0))
+              (let [method-name (subs method-str 1)]
+                (fn [this & args]
+                  (let [^Method method (find-method (class this) method-name (mapv class args))]
+                    (Reflector/prepRet (.getReturnType method) (.invoke method this (into-array Object args))))))))))
```
vlaaad, line 575: Note: `Reflector/prepRet` is an important part of interop because Clojure recognizes only `Boolean/FALSE` and `nil` as falsey, and not any other Boolean instance like `(Boolean. false)`, for example `(if (Boolean. false) :then :else)` returns `:then`. Java method invocations create new Booleans instead of reusing existing `Boolean/TRUE` and `Boolean/FALSE`, and `prepRet` ensures that boolean method invocation results are either `Boolean/TRUE` or `Boolean/FALSE`

kordano: Thanks @vlaaad. I'll have a look today and come back to you.

PR #145, opened 2020-04-30
unifying error handling and introducing logging

### LINKED ISSUE #54: unify error handling
The exceptions should have a unified concept across all namespaces.

TimoKramer:
unifying error handling and introducing logging
    
    To unify error handling I moved the raise macro in the tools
    namespace to access it easily from everywhere. It now is
    capable to be used by clj and cljs. There is an error logged
    everytime the raise macro is called.
    
    For logging I introduced taoensso.timbre to be able to log in
    Clojure and Clojurescript. This library supports both
    structured and unstructured logging and suits our needs for
    log analysis. The macros of this library can be called where
    needed and can process all data structures.
    
    The biggest problem is bug CLJ-865 in Clojure that prevents
    passing line numbers between macros. Luckily timbre is
    awesome and has a low level macro that makes it possible to
    circumvent this bug. After a lot of fiddling it is now
    possible to use a raise macro in the tools namespace and see
    line numbers in the logs.
    
    - Closes #54
    - Great! https://github.com/ptaoussanis/timbre/commit/36785d68715bf9d7496c30eef5e3839c4e45e28e

Squashed commit of the following:

commit f59e83d3c8bd6cdd7ab23644353c7cf4a819c669
Author: Timo Kramer <fw-git@timokramer.de>
Date:   Thu Apr 30 17:32:48 2020 +0200

    set log-level for tests to fatal

commit 35f4cf299921ad49f8de90fef1aee0de3e1f600c
Author: Timo Kramer <fw-git@timokramer.de>
Date:   Thu Apr 30 17:02:59 2020 +0200

    using log directly is best

commit 4d484c6f4541d9cc4e8f5481474ce8eb5a8f7308
Author: Timo Kramer <fw-git@timokramer.de>
Date:   Thu Apr 30 15:55:48 2020 +0200

    transact has special sauce

commit abe0bac6546cf2955de0b7fbf0645843af067ed8
Author: Timo Kramer <fw-git@timokramer.de>
Date:   Wed Apr 29 15:18:15 2020 +0200

    implemented logging and error handling in tools namespace

commit 902411cd59292b683949b69fe49786db7256461f
Author: Timo Kramer <fw-git@timokramer.de>
Date:   Thu Apr 9 20:31:30 2020 +0200

    raise moved to tools namespace


TimoKramer:
possible is to use extra macros in tools-namespace for wrapping the lib like in kabel: https://github.com/replikativ/kabel/blob/master/src/kabel/platform_log.cljc
I could do that.

TimoKramer:
rebased and fixed conflicts. please merge if you think this is fine. @kordano

project.clj
```diff
@@ -10,6 +10,7 @@
                  [org.clojure/tools.reader "1.3.2"]
                  [environ "1.1.0"]
                  [io.replikativ/hitchhiker-tree "0.1.7"]
+                 [com.taoensso/timbre "4.10.0"]
```
kordano, line 13: We should add a section about logging in the readme, so one can set up their timbre config that also covers our errors. Maybe also some hints about having more than just error messages but also error data like `{:type :backend-does-not-exist :config store-config}`.

src/datahike/tools.cljc
```diff
@@ -21,3 +23,10 @@
 (defn get-time []
   #?(:clj (java.util.Date.)
      :cljs (js/Date.)))
+
+(defmacro raise [& fragments]
```
kordano, line 27: What differences are between cljs and clj in timbre?

test/datahike/test.cljc
```diff
@@ -37,6 +37,8 @@
     datahike.test.upsert
     ))
 
+(taoensso.timbre/set-level! :fatal)
```
kordano, line 40: What levels should we use in Datahike?

src/datahike/tools.cljc
TimoKramer: afaik timbre abstracts away the differences. I think in clj it logs via your used logging-impl and in JS it uses console.log/error or something like that.

test/datahike/test.cljc
TimoKramer: I thought it would make sense to set the level in the http-server for example. as a library I would not set anything, but I am open for other opinion. I just set it for testing, do you think this does not make sense?

project.clj
TimoKramer: added documentation

TimoKramer:
does currently not show line number

TimoKramer:
https://github.com/ptaoussanis/timbre/issues/22 currently not possible to log with line number through timbre :(

TimoKramer:
either we accept it like this or we log in the line where we want to log.

whilo:
This is a hard one. I think line numbers are very important information when you try to debug or profile. It is frustrating that the issue in the Clojure compiler is not fixed yet, this is something that makes metaprogramming a lot harder. Can we try to push to get the upstream bug fixed first, e.g. by reaching out to @puredanger?

puredanger:
If you're referring to CLJ-865, we've taken stabs at this multiple times. There is not "a" fix to push and it is not clear if it's even possible to make a change there without breaking existing expectations. If you're interested in us looking at it again, please vote at https://ask.clojure.org/index.php/3838/macroexpansion-discards-form-metadata.

TimoKramer:
I would rewrite this PR to not use `raise`-macro but log and throw on all the places where appropriate. that means duplication but it is not easy to avoid.

src/datahike/tools.cljc
whilo, line 27: We could enforce here in the macro that the first argument to `raise` must be a message string and the second argument must be a map containing a `:type` key.

whilo: I think we should also have a `:level` set by default, which is probably always going to be `:error` (?)

kordano: Yeah, sometimes we don't need all info level notifications.

TimoKramer:
I did my best to write an atom that specs all the logs and an appender that writes into this atom. I couldn't get the spec working:( I think now is a good time to accept that it is like it is and it is too much effort for me to do the specing of the logs. The specing of the raise-macro makes IMHO no sense since there are vars, functions and other stuff that the macro fails on since a macro specs at compile time.
At least we have line numbers now.

Issue #146, opened 2020-05-06
Datahike does not support non-date attributes on tx entities


vlaaad:
### Context
It is useful to annotate transaction entities with information about transactions themselves (see [here](https://docs.datomic.com/on-prem/best-practices.html#add-facts-about-transaction-entity) for more info).

### Problem
Adding non-date attributes to transaction entities breaks filtered databases such as `d/as-of`.

### Repro
```clj
(let [uri "datahike:mem://as-of-with-tx-attrs-repro"
      _ (d/create-database uri :initial-tx [{:db/ident :operation/reason
                                             :db/valueType :db.type/string
                                             :db/cardinality :db.cardinality/one}])
      conn (d/connect uri)
      _ (d/transact conn [{:db/id "datomic.tx"
                           :operation/reason "Reason for this tx”"}])]
  (d/as-of (d/db conn) #inst "2020"))
```

### Expected output

I would expect to receive a db as-of that time point

### Actual output
An exception is thrown:
```clj
{:via [{:type clojure.lang.ExceptionInfo
        :data {:clojure.error/phase :print-eval-result}
        :at [clojure.main$repl$read_eval_print__9086 invoke "main.clj" 442]}
       {:type java.lang.ClassCastException
        :message "class java.lang.String cannot be cast to class java.util.Date (java.lang.String and java.util.Date are in module java.base of loader 'bootstrap')"
        :at [datahike.db$filter_as_of_datoms$as_of_pred__39689 invoke "database_connection.clj" 84]}]
 :trace [[datahike.db$filter_as_of_datoms$as_of_pred__39689 invoke "database_connection.clj" 84]
         [datahike.db$filter_txInstant$fn__2414 invoke "db.cljc" 463]
         [clojure.lang.ArrayChunk reduce "ArrayChunk.java" 63]
         [clojure.core.protocols$fn__8154 invokeStatic "protocols.clj" 136]
         [clojure.core.protocols$fn__8154 invoke "protocols.clj" 124]
         [clojure.core.protocols$fn__8114$G__8109__8123 invoke "protocols.clj" 19]
         [clojure.core.protocols$seq_reduce invokeStatic "protocols.clj" 31]
         [clojure.core.protocols$fn__8146 invokeStatic "protocols.clj" 75]
         [clojure.core.protocols$fn__8146 invoke "protocols.clj" 75]
         [clojure.core.protocols$fn__8088$G__8083__8101 invoke "protocols.clj" 13]
[...]

PR #152, opened 2020-05-15
Support non date tx attributes

vlaaad:
This is an attempt to solve #146 

I don't fully understand the internals of Datahike, so if this PR doesn't make sense, feel free to reject it!

The way I see it, `filter-txInstant`'s purpose is to produce a set of tx ids that satisfy a predicate taking instant of a transaction. The problem was that it assumed transactions only have `:db/txInstant` attributes. I added a check so only `:db/txInstant` attr is used, and all other attributes are skipped. 

I also used transducers to improve performance.


kordano:
Great, thank you @vlaaad

PR #154, opened 2020-05-16
153 safe guard query engine

### LINKED ISSUE #153: Support safe guarded queries without external calls
Currently the query engine does function name resolution and Java reflection to find methods. This allows script injection in cases where we want to openly expose the query endpoint.

whilo:
This pull request closes #153 and, assuming ithere are no further loop holes, allows to run the query engine publicly, e.g. through the REST endpoint of `datahike-server`.


whilo:
I thought about it, we can solve this with static analysis on the query independent of Datahike, no need for runtime checks. A good reason to reactivate Datopia.

PR #159, opened 2020-05-24
Query now supports returning maps

### LINKED ISSUE #149: Query does not support returning maps
### Problem
Datomic query engine can [return maps](https://docs.datomic.com/on-prem/query.html#return-maps) instead of tuples by specifying `:keys`, `:strs` or `:syms` clauses. [Datomic Differences](https://github.com/replikativ/datahike/blob/master/doc/datomic_differences.md) article says that Datahike has no differences with `q`, but it does not support return maps.

### Repro

```clj
(d/q '[:find ?e :keys entity :in ?e] "entity")
```

### Expected output

```clj
#{{:entity "entity"}}
```

### Actual output

```clj
#{["entity"]}
```
Alternatively, the Datomic Differences article could state that it does not support return maps — it will help people migrating from Datomic to Datahike to see what to watch out for.

### Workaround
You can post-process query results with `zipmap` to get expected result:
```clj
(->> (datahike/q '[:find ?e :in ?e] "entity")
     (map #(zipmap [:entity] %))
     set)
```

I find return maps very convenient for query post-processing since using keywords on maps is more meaningful than using `nth` on tuples.

TimoKramer:
### Query now supports returning maps

Since mid 2019 Datomic supports returning maps and since Datahike
wants to map Datomics Query-API I added it. The most changes were
made on the datalog-parser:
https://github.com/lambdaforge/datalog-parser/pull/9

The changes made to datahike are a function convert-to-return-maps
that takes a set and maps the given keys on its values and returns
a vector of the created maps. The changes made to the function q
are - besides formatting - a check if returnmaps are parsed and if
so calling aforemention function to convert the result.

Besides adding another parse-step and the check if there are
returnmaps parsed there should not be too much of a performance
impact as far as I can see it. The behaviour should now map
Datomics.

- Closes #149
- Depends on https://github.com/lambdaforge/datalog-parser/pull/9

Thanks to vlaad for bringing this up

### DIFF
--- a/src/datahike/query.cljc
+++ b/src/datahike/query.cljc
@@ -160,8 +160,8 @@
              (reduce (fn [acc t2]
                        (conj! acc (join-tuples t1 idxs1 t2 idxs2)))
                      acc (:tuples rel2)))
-           (transient []) (:tuples rel1)))
-       ))))
+           (transient []) (:tuples rel1)))))))
+
 
 ;; built-ins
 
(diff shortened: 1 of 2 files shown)


TimoKramer:
CI failing because of datalog-parser-dependency.

Issue #167, opened 2020-05-27
Automated Release Process


TimoKramer:
Since we want to release Datahike more often there should be an automated release process. Which includes:

- [x] unittesting clj
- [x] integrationtesting with all the official backends in-memory, file, PG/JDBC
- [x] integrationtesting with all the official backends LevelDB
- [x] release to Clojars
- [x] release to maven
- [x] central and testable CI config
- [x] documentation 
- [ ] testing cljs
- [ ] performancetesting
- [ ] compliance tests for backends
- [ ] automated tests for replikativ libraries and other products
- [x] datahike-server releases to docker hub

We should also provide a guideline in terms of testing to unofficial/developing backends with an example to use GitHub Actions of CircleCI. We should also provide automated testing for all the libs that are under our control that are used by datahike and probably others.

whilo:
For external backends we should include compliance tests. We already have a version of this in konserve, but it is not done yet. Basically all backends should have to run through the same comprehensive test suite.

TimoKramer:
I personally find it neat to have a CI building and testing my code on a git push. that is already working.

Following that principle it would be convenient to tag a commit and push that tag to github and that would create an artifact and release it on clojars. There was the wish to have SNAPSHOT-releases be released on commits to development-branch and have real releases be built and released on commit on master.

TimoKramer:
storing secrets in circleci is necessary to deploy to clojars: https://circleci.com/docs/2.0/security

TimoKramer:
I created a first draft to try out a semi automated release model here: https://github.com/TimoKramer/datahike/commit/623808ec8ab517a4008b29af8d7681e13a677ff7
please ask for access to the repo if you want to try it. you can release a '.*-SNAPSHOT'-named artifact to clojars when committing on development and you can release a not '.*-SNAPSHOT'-named release when committing on master.

TimoKramer:
tagging should be automatic. need to find a way to do that without uploading a write-key to circleci. local git-hook might be possible.

TimoKramer:
The env vars LEIN_USERNAME and LEIN_PASSWORD shoud be set inside the clojars-deploy context. and removed from the projects.
The env var CIRCLE_TOKEN should be set inside the orb-deploy context.

TimoKramer:
Closing for now and coming back when CLJS-testing is in work

PR #177, opened 2020-06-04
167 automated release process

TimoKramer:
### Integration Tests for Datahike (#167)                                
                                                                      
 On the way to a fully automated test and release cycle               
 I introduced the first batch of integration tests and                
 implemented them into the circleci test run. Since                   
 Kaocha was already a dev dependency I am using it                    
 to run the integration suite.                                        
                                                                      
 "lein kaocha integration" is the command to run the                  
 integration tests and they are located in the foler                  
 test/datahike/integration_test/. All the files                       
 located there will be run by the integration tests                   
 but not by the common tests run with "lein test".                    
                                                                      
 I added a folder bin for running the integration tests.              
 Also there is the script to use when running other                   
 kaocha functions named "kaocha". Furthermore there will              
 soon be a script that starts the release-process when I              
 will not find a nice way to do it via circleci.                      
                                                                      
 Tests will be run on push and PR to master and                       
 development.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,30 +1,37 @@
-version: 2 # use CircleCI 2.0
-jobs: # basic units of work in a run
-  build: # runs not using Workflows must have a `build` job as entry point
-    working_directory: ~/datahike # directory where steps will run
-    docker: # run the steps with Docker
-      - image: circleci/clojure:lein-2.8.3 # ...with this image as the primary container; this is where all `steps` will run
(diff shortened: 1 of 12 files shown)


TimoKramer:
the return-maps-PR should be merged first.

TimoKramer:
image
I wonder why it shows changes to src/query.cljc and test/query.cljc. there should  be no change.

kordano:
Nice work, thanks.

PR #191, opened 2020-06-23
Implemented automated releases to Clojars

TimoKramer:
Since our goal is to release more often I implemented a first draft
to automate the releases. Releases are a very individual thing and
I tried to cover datahikes previous manual process as far as
possible. This means that a commit to development creates a release
if there is a SNAPSHOT-version specified in project.clj and a
commit to master creates a release if there is a non-SNAPSHOT-version
specified in project.clj.

Besides that there is some streamlining of the steps happening in
the usual testing and building procedure. Furthermore I introduced
the new workflow-feature of CircleCI and now the steps are able to
run concurrently. This seems to be quite a bit faster.

The decision is to keep it simple but avoiding manual work in the
future. Therefor I am sticking with CircleCI even if there are
other good options. Another decision is to continue keeping the
current version in project.clj. This has downsides since you can
not easily tag a commit with git and get a new release because you
would have to adjust the version manually in project.clj and tag
it. This is error prone.

Either tagging continues to be manually or it is solved with a
git hook. That way is described in doc/release.md. It is not the
nicest way to do it but I did not want to use a leiningen plugin
or upload a write-key to CircleCI. There should be sufficient
documentation in the release.md.

- Refers https://github.com/replikativ/datahike/issues/167

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,37 +1,137 @@
-version: 2.0
-jobs:
-  build:
-    working_directory: ~/datahike
+version: 2.1
 
+executors:
+  leiningen:
     docker:
-      - image: circleci/clojure:lein-2.9.3
+      - image: circleci/clojure:openjdk-8-lein
+    environment:
+      LEIN_ROOT: nbd
+      JVM_OPTS: -Xmx3200m
+    working_directory: /home/circleci/datahike
+
+jobs:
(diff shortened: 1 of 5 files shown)


TimoKramer:
the CI failed because it can not deploy a replikativ-release to clojars because it does not have permission.

PR #192, opened 2020-06-24
hashes are not correct when reconnecting

kordano:


TimoKramer:
who should review?

whilo (dismissed):
This looks good to me. But we really need to solve the indentation issue.

jsmassa (approved):
Good job!

PR #201, opened 2020-07-10
62 upsert and transact performance improvement

grischoun:
Improves transaction performance by rewriting upserting related code in Datahike and by adding an upsert operation to the Hitchhiker tree.


grischoun:
Some remarks:

- Tests will not pass until a new version of the Hitchhiker tree with the upsert operation is deployed.
- One test is still not passing. It is about 'listen!'. We need to have a discussion about the semantics of 'listen!'.

project.clj
```diff
@@ -10,7 +10,7 @@
                  [org.clojure/tools.reader "1.3.2"]
                  [environ "1.2.0"]
                  [com.taoensso/timbre "4.10.0"]
-                 [io.replikativ/hitchhiker-tree "0.1.7"]
+                 [io.replikativ/hitchhiker-tree "0.1.7"] ;; TODO: Change to the version of the hh-tree that has the upsert op.
```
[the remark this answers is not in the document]
whilo: This will require the definition of the deserialization bits for konserve as well: https://github.com/replikativ/hitchhiker-tree/blob/master/src/hitchhiker/tree/bootstrap/konserve.cljc#L142

src/datahike/db.cljc
```diff
@@ -1149,19 +1149,72 @@
...
+           :attribute (.-a datom)
+           :datom     datom})))))
+
+
+(defn- with-datom-upsert [db ^Datom datom]
+  (validate-datom-upsert db datom)
+  (let [indexing?      (indexing? db (.-a datom))
+        schema?        (ds/schema-attr? (.-a datom))
+        keep-history?  (and (-keep-history? db) (not (no-history? db (.-a datom))))
+        db             (if (or keep-history? indexing?)
```
whilo, line 1169: This means that when we want to keep the history or index we still use `-search` and B+-tree lookups here. Have you thought about what would be necessary to avoid that as well? This is necessary because we need `v` and `t` for the Datom, right? Maybe another type of UpsertOp could extract this once it arrives in the leaf and only insert itself (not deleting `old`)?

project.clj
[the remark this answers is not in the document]
grischoun: > Why do you need the upsert op in the hitchhiker-tree repo? We should be able to define it in Datahike and I think it makes sense, because it contains bits that are not relevant on the hitchhiker-tree level. We should definitely reference it from the hitchhiker-tree documentation though.

Right this is not needed. Using hh-tree's  'enqueue' instead now.

src/datahike/index/persistent_set.cljc
```diff
@@ -46,5 +46,8 @@
 (defn -insert [set datom index-type]
   (set/conj set datom (index-type->cmp-quick index-type)))
 
+(defn -upsert [set datom index-type]
```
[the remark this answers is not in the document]
grischoun: Fixed.

grischoun (commented):
Done.

test/datahike/test/upsert_implem_test.cljc
```diff
@@ -37,17 +37,17 @@
         (let [new [5 :name "Jo" 3]
               tree-after (upsert-helper tree new)]
           ;; Is new in?
-          (is (= new (msg/lookup tree-after new)))))
+          (is (= [new nil] (first (msg/lookup-fwd-iter tree-after new))))))
```
whilo, line 40: Why do you have to use the iterator now?

src/datahike/index/hitchhiker_tree/upsert.cljc
```diff
@@ -26,7 +26,7 @@
   (-affects-key [_] key)
   (-apply-op-to-coll [_ kvs]
     (-> (or (remove-old kvs key (partial dissoc kvs)) kvs)
-      (assoc key value)))
+      (assoc key nil)))
```
whilo, line 29: Can you remove the value also the UpsertOp record itself?

src/datahike/db.cljc
```diff
@@ -1193,19 +1193,63 @@
...
+
+(defn- with-datom-upsert [db ^Datom datom]
+  (validate-datom-upsert db datom)
+  (let [indexing?     (indexing? db (.-a datom))
+        schema?       (ds/schema-attr? (.-a datom))
+        keep-history? (and (-keep-history? db) (not (no-history? db (.-a datom)))
+                        (not= :db/txInstant (.-a datom)))]
+    (cond-> db
+      ;; Optimistic remove of the schema entry
+      schema? (try
```
whilo, line 1216: What is the benefit of this optimistic update?

grischoun: In the end the goal is to remove the schema entry old value if there is one. But we don't know if there is an old version or not, so we try to remove it anyway and if it fails then we just go on.

In the old code we had the following to check whether there is an old version of the schema entry or not.
`(if-some [removing ^Datom (first (-search db [(.-e datom) (.-a datom) (.-v datom)]))]`
But this search is expensive and it is what we wanted to avoid in the first place.

src/datahike/index/hitchhiker_tree/upsert.cljc
grischoun: Done.

test/datahike/test/upsert_implem_test.cljc
grischoun: Because `lookup`returns the value and not the key, and with our latest change we are not storing the value anymore, so I can no longer use `lookup`.

src/datahike/index/hitchhiker_tree/upsert.cljc
grischoun: Ooppps, no, it is not done yet.

grischoun: Finally done now.

src/datahike/db.cljc
whilo: Ok, that makes sense. Thanks.

whilo (approved):
This is very cool! Thanks a lot for putting in the effort.

whilo (approved):

PR #210, opened 2020-07-25
Add small benchmark utility

jsmassa:
First draft; still need to think about how it is supposed to be executed and about output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ release.clj
 .#*
 /.gitold/COMMIT_EDITMSG
 .cpcache
+lsp/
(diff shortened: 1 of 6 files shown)


kordano:
Looks good to me. We would add test namespaces later for compliance. E.g. that transactions for 100K should last longer than X seconds.

TimoKramer (approved):

jsmassa:
I will add some documentation to the README

TimoKramer (approved):

TimoKramer:
thanks @jsmassa for the docs! Ready to merge fmpov.

PR #222, opened 2020-08-26
Configure database name with environ

### LINKED ISSUE #217: configuring database name
for running datahike with environment variables there is no way to configure the name of the database(s). There needs to be an option to configure database name like `DATAHIKE_NAME=labrador`

TimoKramer:
For a use case where there are multiple databases running like in
datahike-server there it is useful to have named databases to
specify the database you want to use in each operation. This
feature was added but there was no way to specify the names in
using environment or others, only via config file. I added the
option to use environ to specify names equal to the other config
values.

Besides that I set the logging of the config to debug-level
because it is logging each time the config is called for exists?,
 create and connect and each time maybe with a different name.
That is confusing when name is not set and I want it to not
confuse users. Documenation is adapted.

- Closes #217

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -28,6 +28,7 @@ That means passing a config as argument overwrites java system properties and us
          :path     nil         ;string
          :host     nil         ;string
          :port     nil}        ;int
+ :name (generated)             ;string
  :schema-flexibility :write    ;keyword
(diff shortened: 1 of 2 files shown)


jsmassa (approved):
Looks good to me

PR #223, opened 2020-08-28
Switched the type to store the entity id in a Datom from `int` to `long`

### LINKED ISSUE #221: long instead of int for entity ids
First of all thanks a lot for starting Datahike and working towards an open-source Datomic-like database :+1: 

I've done a few experiments with Datahike and tried to import some real world data from Datomic. I noticed that Datahike is still using `int` as type for the entity id in `datahike.datom/Datom`. Would it make sense to switch to `long`?

Best regards,

Max

maxweber:
This will probably cause breakage for existing Datahike installations. Therefore this PR is only a starting point for a discussion, if this type switch is reasonable or not. The main reason for the switch is to be able to important existing Datomic data into Datahike.

I'm happy to do the other required tasks to avoid breakage or write a migration tool. But I would need some guidance to figure out what tasks are necessary :sweat_smile:

Closes #221

### DIFF
--- a/src/datahike/datom.cljc
+++ b/src/datahike/datom.cljc
@@ -10,7 +10,7 @@
   (datom-tx [this])
   (datom-added [this]))
 
-(deftype Datom #?(:clj  [^int e a v ^long tx ^:unsynchronized-mutable ^int _hash]
+(deftype Datom #?(:clj  [^long e a v ^long tx ^:unsynchronized-mutable ^int _hash]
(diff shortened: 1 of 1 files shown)


whilo:
I think this makes sense. @kordano We have inherited this constraint from DataScript, right? Since we aim to not break compatibility with Datomic unless we have a good reason to, we should probably fix this soon.

kordano:
Introduced with release  `0.4.1485`.

PR #232, opened 2020-09-20
Versioning and garbage collector for persistent-sorted-set backends.

whilo:
This pull request implements garbage collection for hitchhiker-tree based databases.


whilo:
I just realized that the tests still require an assertion for the history index.

kordano:
Should the GC function be available to the user? If so, then some documentation to the README needs to be added.

test/datahike/test/store.cljc
```diff
@@ -51,3 +52,49 @@
...
+(deftest test-hitchhiker-tree-gc
+  (testing "Testing gc after appending data."
+    (let [uri "datahike:file:///tmp/datahike-hh-gc-test-fs"
+          _   (d/delete-database uri)]
+      (is (not (d/database-exists? uri)))
+      (d/create-database uri :schema-on-read true)
+      (let [conn (d/connect uri)
+            now  (fn [] (java.util.Date.))]
+        (d/transact conn (vec (for [i (range 1 1001)]
+                                { :db/id i, :name "Ivan", :age 15 })))
```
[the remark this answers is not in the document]
whilo: Should be fixed with the new tests.

src/datahike/connector.cljc
```diff
@@ -15,52 +16,155 @@
...
+                       :rschema         rschema
+                       :system-entities system-entities
+                       :ident-ref-map   ident-ref-map
+                       :ref-ident-map   ref-ident-map
+                       :config          config
+                       :meta            meta
+                       :hash            hash
+                       :max-tx          max-tx
+                       :max-eid         max-eid
+                       :op-count        op-count
```
jsmassa, line 148: Could be made more condensed by using `select-keys` on db-after and merge it with the changes. I am just saying because you are refactoring this anyway in this PR.

test/datahike/test/gc_test.cljc
```diff
@@ -0,0 +1,141 @@
...
+      (is (thrown-msg? "Index needs to be properly flushed before marking."
+                       (mark (:eavt
+                              (:db-after
+                               (d/with @conn [{:db/id 100
+                                               :age   5}]))))))
+
+      ;; check that we can still read the data
+      (d/transact conn txs)
+      (<?? S (gc! @conn (Date.)))
+      (is (= 1000 (d/q count-query @(d/connect cfg)))))))
```
jsmassa, line 64: Instead of the `testing` statement at the top it would be more helpful to have several `testing` statements throughout the test describing what is being tested so problems are better visible on the test report. Here, it looks to me like the comments could be largely turned into `testing` statements enclosing the statements that they are concerned with.
A top level testing statement is usually not needed when the name of the test is chosen well.

jsmassa (commented):
Thank you for all the work you put into this and that you provided documentation as well. Very nice! 

I think I understand why you implemented the versioning together with the garbage collector, it would have been better though to have the versioning added with a separate PR. The fact that it includes the two features should be at least be reflected in the PR title so it will appear in the Changelog of the Github release. 
It also looks to me like the GC is not only available for the persistent set, but also for the hitchhiker-tree. If that is correct then this should be also corrected in the PR title.

Other than that, your tests could use a bit more structure so that tracing problems becomes easier.

test/datahike/test/gc_test.cljc
whilo: Agreed and fixed.

src/datahike/connector.cljc
whilo: This is possible, but it also requires to transform the result of `select-keys` into a map, which makes the code a bit convoluted. I think in this case the number of lines saved is not worth the tradeoff of easy readability.

whilo:
I looked into providing `gc` also for the hitchhiker-tree. I had implemented `mark` for it as well, but currently there is no easy way to detect that its nodes are not dirty after the initial `connect` call, so it would require fixing. I am not for throwing out the hitchhiker-tree right away, since it forces us to keep the interfaces general, but I would prefer spending my time not on adding features to it and instead on moving Datahike forward.

src/datahike/connector.cljc
```diff
@@ -34,7 +34,8 @@
   (let [{:keys [eavt aevt avet temporal-eavt temporal-aevt temporal-avet
                 schema rschema system-entities ident-ref-map ref-ident-map config
                 max-tx max-eid op-count hash meta store]} db
-        backend (di/konserve-backend (:index config) store)]
+        backend (di/konserve-backend (:index config) store)
+        in-memory? (not= :mem (-> config :store :backend))]
```
jsmassa, line 38: `in-memory?` is exactly the opposite from what is being tested here, so this really has to be renamed

whilo: Thanks, that was indeed precision in the wrong direction :smile: .

jsmassa (approved):

PR #236, opened 2020-10-06
211-attr-refs

jsmassa:
Open questions:
- flexibility or fixed expectation of ident (keyword) in functions like
  - is-attr
  - validate-attr
  - search-indices
- replace idents in schema cache with refs (and rschema)?
- naming convention to differentiate between attribute as in database and ident of attribute


test/datahike/test/attribute_refs/query.cljc
```diff
@@ -0,0 +1,242 @@
...
+
+(deftest test-q-in
+  (let [entities [{:db/id 1, :name "Ivan", :age 15}
+                  {:db/id 2, :name "Petr", :age 37}]
+        db (d/db-with ref-db (shift-entities ref-e0 entities))
+        query '{:find [?e]
+                :in [$ ?attr ?value]
+                :where [[?e ?attr ?value]]}
+        ref (fn [key] (get-in db [:ident-ref-map key]))]
+    (println "name" (ref name))
```
whilo, line 61: You missed a println ;).

test/datahike/test/query.cljc
```diff
@@ -117,24 +117,24 @@
                (d/db-with [{:db/id 1, :name "Ivan", :age 15}
                            {:db/id 2, :name "Petr", :age 37}
                            {:db/id 3, :name "Ivan", :age 37}]))]
-    (testing "Relation binding"
```
whilo, line 120: Why are these tests (and others) commented out?

whilo (commented):
Nice work! I am not familiar with the details, so I will not approve for now and take a closer look again after somebody else has done a review as well.

test/datahike/test/query.cljc
jsmassa: just forgot to uncomment

jsmassa:
The new changes are merged now, but we still have to clear if https://github.com/replikativ/datahike/blob/0bf89ab859e326ed6c6cec9a77118b5a703b1198/src/datahike/db.cljc#L964
is correct as the filtering is not in the dev branch anymore. To this point I am convinced that the filtering was overlooked at some point and it was never noticed since the data needed is still in the :avet index, there's just more being saved than intended.

kordano:
Yes, I had a look at the history and somehow it was overlooked. Thanks for recognizing that. Also could you have a look at the confilcts again.

kordano (approved):

PR #248, opened 2020-10-23
215 qseq support

TimoKramer:
draft of qseq support with problem using paginate


TimoKramer:
when using qseq, should the result be exactly the same as q but with sequence? or is it ok to have different sorting and therefore different results in terms of eg. offset and limit. @kordano @whilo

TimoKramer:
won't merge this because there is more architectural decisions to make.

PR #251, opened 2020-11-03
Add Tuples

grischoun:


src/datahike/db.cljc
```diff
@@ -1375,277 +1495,339 @@
...
+                  (retry-with-tempid initial-report report initial-es e upserted-eid)
+                  (let [eid (or upserted-eid allocated-eid (next-eid db))]
+                    (recur (allocate-eid report e eid) (cons [op eid a v] entities)))))
+
+              (and (ref? db a) (tempid? v))
+              (if-let [vid (get tempids v)]
+                (recur report (cons [op e a vid] entities))
+                (recur (allocate-eid report v (next-eid db)) es))
+
+              (and (homogeneous-tuple-attr? db a)
```
grischoun, line 1688: From here to `(= op :db/add)` are changes that I have made.

test/datahike/test/tuples_test.cljc
```diff
@@ -0,0 +1,423 @@
...
+                {:db/id 4 :a "A" :b "B" :c "c"}
+                {:db/id 5 :a "a" :b "b" :c "C"}
+                {:db/id 6 :a "A" :b "b" :c "C"}
+                {:db/id 7 :a "a" :b "B" :c "C"}
+                {:db/id 8 :a "A" :b "B" :c "C"}]))]
+    (is (= [6]
+          (mapv :e (d/datoms db :avet :a+b+c ["A" "b" "C"]))))
+    (is (= []
+          (mapv :e (d/datoms db :avet :a+b+c ["A" "b" nil]))))
+    (is (= [8 4 6 2]
```
TimoKramer, line 420: Is it correct that 4 and 2 are returned here as well? what are the matching rules here? I would have expected to only get back [8 6]

test/datahike/test/tuples_test.cljc
```diff
@@ -0,0 +1,423 @@
...
+          (mapv :e (d/index-range db :a+b+c ["A" "B" "C"] ["A" "b" "c"]))))
+    (is (= [8 4]
```
TimoKramer, line 422: same here

TimoKramer (commented):
How about some documentation on this new functionality?

src/datahike/db.cljc
```diff
@@ -1375,277 +1495,339 @@
...
+                (> (count v) 8))
+              (raise "Cannot store more than 8 values for homogeneous tuple: " entity
```
TimoKramer, line 1690: Probably it suffices to send the corresponding entity in the attached map?

src/datahike/db.cljc
```diff
@@ -1375,277 +1495,339 @@
...
+              (raise (str "Cannot store heterogeneous tuple: expecting " (count (-> db -schema a :db/tupleTypes)) " values, got " (count v))
+                {:error :transact/syntax, :tx-data entity})
+
+              (and (heterogeneous-tuple-attr? db a)
+                (not (apply = (map s/valid? (-> db -schema a :db/tupleTypes) v))))
+              (raise (str "Cannot store heterogeneous tuple: there is a mismatch between values " v " and their types " (-> db -schema a :db/tupleTypes))
+                {:error :transact/syntax, :tx-data entity})
+
+              (and (not (::internal (meta entity)))
+                (composite-tuple-attr? db a))
```
TimoKramer, line 1714: I'd love to see this the other way round just for the sake of consistency with the above blocks.

TimoKramer:
```
(and (composite-tuple...
    (not (::internal...
```

src/datahike/db.cljc
TimoKramer: but on the other hand it is consistent with some other raises. I don't know... I would probably leave it to the map but you can decide.

TimoKramer:
Seems to me that this is the commit that introduced all the reformatting: https://github.com/replikativ/datahike/commit/9fd3d24c042e62a925e671633f48988889baeeb9

test/datahike/test/tuples_test.cljc
grischoun: It is correct because "A" is smaller than "a" in Java/Clojure (e.g.: `(.compareTo "B" "b") => -32`).
So `["A" "b" "C"]` is smaller than `["A" "b" "c"]`.

test/datahike/test/tuples_test.cljc
grischoun: Also same reason as above.

grischoun:
@kordano The tuples branch is ready again.

TimoKramer:
How about a squash?
Just do something like 
```
git checkout development
git pull
git checkout -b temp
git merge --squash tuples
git commit
```

grischoun:
> How about a squash?
> Just do something like
> 
> ```
> git checkout development
> git pull
> git checkout -b temp
> git merge --squash tuples
> git commit
> ```
> 
> When you're fine with the changes then force push it to the tuples branch.

Done

TimoKramer (approved):

PR #257, opened 2020-11-26
Documentation

jsmassa:
Pull-API documentation for core developers


doc/development/pull-api.md
```diff
@@ -0,0 +1,79 @@
...
+     - attribute, :as, new name
+     - e.g. [:aka :as :alias]
+            [:name :as "Name"]
+
+## Pattern = (Pull-)Spec
+
+Record with keys
+- :wildcard? = boolean value indicating if wildcard expansion has to be done
+- :attrs = map of attribute names as in selector and map with keys
+           - :attr = attribute name as in database; only different from key when key describes reverse attribute
```
TimoKramer, line 47: this seems not to render correctly as a nested list

TimoKramer: maybe not indenting that far?

doc/development/pull-api.md
```diff
@@ -0,0 +1,79 @@
+# Pull-API: Terminology
+
+
+## Selector
+
+Quoted vector holding one or several of the following expressions:
+1) wildcard
+   - pulls all datoms for entity id
+   - e.g. * -> {:db/id 1 :name "Petr" :child [{:db/id 2}]}
```
TimoKramer, line 9: with some backticks you can get `monospace` for code snippets here.

TimoKramer (commented):
is the development-folder necessary here? will this work fine with cljdoc? and **using** the datahike-api is not **development for** datahike, so this is mistakable.

As an unexperienced reader I would love to see some examples here that show how to use the snippets. Maybe just one or two are sufficient. But examples are best for someone that has no idea on where to put these snippets.

A link to the documentation of the cljdoc api-namespace for the pull api might help.

TimoKramer:
image

doc/development/pull-api.md
jsmassa: Thanks for the tip, I will adjust the formatting

doc/development/pull-api.md
jsmassa: I see... I wanted to ave the complete explanation at the right side of the equality sign, but apparently that is not possible with pure markdown like that

jsmassa:
Maybe I should have made this more clear, but I did not write this documentation for people using the Pull-API, but for people like us trying to refine some aspects of the code in the pull-api namespace and therefore having to understand it first. That is why I put it in a development folder. 
An end user of the Pull-API will neither work with frames nor with patterns, but only with selectors. 
But I guess adding a few examples and linking the doc doesn't hurt in any case, so I will get to that.

TimoKramer:
Can you put this above description below the header of the doc to tell people who this is for? That would be helpful I think.

TimoKramer (dismissed):
I think this is good, as far as I can see it. But adding a description on who this is addressing would be good :+1:

TimoKramer (approved):

PR #265, opened 2020-12-10
Introduce temporary fix for predicates

jsmassa:
Only temporary since order is still relevant, but now the clauses get reordered to make it work.
The query engine should be made intelligent in the future.


whilo:
I think this is not a good idea, because predicates have a filtering effect on set size and by doing them only after grounding everything first you can significantly hit query performance in a large portion of queries that have predicates. Query planners generally push them to the top to reduce sets to join over as much as possible. We also should not reorder queries clauses automatically, but rather do so through an opt-in procedure or with the help of diagnostics tool for now.

jsmassa:
I agree that it is not a good solution, but it is a quick one to at least make the database retrieve the correct data set and not leave things in the set that should have been filtered out. I know very well why I only called it a "temporary fix".

So far, the query engine only applies the predicate to the set of tuples already retrieved from the database and will not be applied to data retrieved later through a pattern clause at a position below said predicate clause. This is why putting the predicates at the end will fix the problem. 

Since we are planning to overhaul the query engine soon anyway, to me it seemed to be an adequate solution for now since I couldn't think of anything that wouldn't require big chunks of the query engine to be rewritten.

PR #266, opened 2020-12-14
Fixed a bug in the Datom's `entryAt` implementation

green-coder:
Added the missing `.` in a constructor call.

### DIFF
--- a/src/datahike/datom.cljc
+++ b/src/datahike/datom.cljc
@@ -27,6 +27,10 @@
        ISeqable
        (-seq [d] (seq-datom d))
 
+       IFn
+       (-invoke [d k] (val-at-datom d k nil))
+       (-invoke [d k v] (val-at-datom d k v))
+
        ILookup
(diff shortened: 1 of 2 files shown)


green-coder:
It would be a good idea to add some tests, I can do that after work or you can also do it on your side:

`(select-keys (dd/datom 1 :foo "bar") [:e :a :v])`

whilo (approved):
Thank you! This looks good to me.

whilo:
@green-coder Adding a test would definitely be very good.

green-coder:
I added some tests.

green-coder:
This PR is ready for reviews.

kordano (approved):

green-coder:
@kordano I just noticed that the namespace for the test file is wrong.

kordano:
Could you create a new PR with the fixed name?

green-coder:
I can do that this weekend.

TimoKramer:
I am fixing this together with the formatting. So no need to do that.

PR #268, opened 2020-12-15
Move config log and avoid password logging

### LINKED ISSUE #263: Improve the config debug logging
This line prints the whole config, including the password that is part of the config when using the JDBC backend.

https://github.com/replikativ/datahike/blob/11b6be4fd4e36ec77500839ffff8a0873f16053a/src/datahike/config.cljc#L127

The line could be changed to:
(log/debug "Using config " (update-in merged-config [:store] dissoc :password))

Or is there a better way to supply the password?

TimoKramer:
- Move the log because it gets logged multiple times currently
- JDBC requires password configuration and this should not be logged
- Closes #263


TimoKramer:
Fixed the transaction logging also with this PR

kordano (approved):

PR #270, opened 2021-01-01
Allow `:db/retract` to be used without specifying a value (Fix #241)

### LINKED ISSUE #241: `:db/retract` without value fails
Datomic (on-prem)'s `:db/retract` has the following [very convenient behavior](https://docs.datomic.com/on-prem/transactions.html#retracting-data):

> To retract data from an existing entity, build a transaction using :db/retract, an existing entity id or entity identifier, the attribute, and optionally value being retracted. If a value is not provided, **all values for the provided entity and attribute will be retracted**.

This is currently not supported in Datahike. Reproducible example:

```
(d/transact conn [{:db/ident :universe
                   :answer   42}])

;; This doesn't work
(d/transact conn [[:db/retract :universe :answer]])
;; => clojure.lang.ExceptionInfo: Cannot store nil as a value at [:db/retract :universe :answer]

;; This works
(d/transact conn [[:db/retract :universe :answer 42]])
```

ema-fox:
`[:db/retract e a]` is not simply an alias for `[:db.fn/retractAttribute e a]` because the latter also retracts any referenced component attributes.

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -1758,13 +1758,13 @@
 
               (= op :db/retract)
               (if-some [e (entid db e)]
-                (let [v (if (ref? db a) (entid-strict db v) v)]
-                  (validate-attr a entity db)
-                  (validate-val v entity db)
-                  (if-some [old-datom (first (-search db [e a v]))]
-                    (recur (transact-retract-datom report old-datom) entities)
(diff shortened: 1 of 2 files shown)


kordano (approved):

kordano:
Thanks for the addition @ema-fox.

PR #271, opened 2021-01-05
267 error on removing schema

grischoun:
Fixes the log error produced when upserting a schema entry.

The solution uses 'throw' only instead of the 'raise' macro, which does both log and throw.


grischoun:
@kordano For a reason I don't understand, it was impossible to add you as a reviewer on this PR.

whilo (approved):
This looks good to me.

TimoKramer (approved):
Thanks!

kordano (approved):

PR #275, opened 2021-01-07
Add predicate var binding check and test

jsmassa:
In order to avoid erroneous results when predicate clauses precede all clauses that bind vars, an error is thrown and the query is being rejected. Would still be better to introduce an (optional) automatic reordering in a future release

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -49,4 +49,7 @@
                    :main-opts ["-m" "hf.depstar.jar" "replikativ-datahike.jar"]}
 
            :format {:extra-deps {cljfmt/cljfmt {:mvn/version "0.7.0"}}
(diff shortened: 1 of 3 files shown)


whilo:
I do not want to be a pain in the ..., but I think it would be beneficial to pull out all static checks from the query runtime code and incorporate them either into the Datalog parser or at least in a separate namespace where we do static analysis that is done independently of running the query. The more we intertwine these steps with the runtime the harder it might be later to pull them out. I am ok with merging this and pulling it out later if it is very pressing though.

whilo (approved):
I think this is fine and we can factor it out as a static analysis pass afterwards.

PR #280, opened 2021-01-28
Add tests and make functions public

jsmassa:
To facilitate further development, here some more of the functions in the pull-api.ns have been make public and tests have been written for them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,5 @@ release.clj
 lsp/
 .calva/
 out/
+.clj-kondo
+.lsp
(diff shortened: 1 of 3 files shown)


groundedSAGE:
Best to use Record Constructors for the Pullspec. This will make the ClojureScript rebase easier. These seem similar and could perhaps be turned into a function similar to the refactoring done here. To solve the same issue.

https://github.com/groundedSAGE/datalog-parser/blob/update-cljs-support/test/datalog/parser_test.cljc#L10

groundedSAGE (approved):
I've added PullSpec constructors which will be required for ClojureScript support.

whilo:
@jsmassa Nice work, you are effectively working towards a specification of the pull API semantics. I think private functions in Clojure are in general not really good, they just are a way to not to commit to a public API for a namespace. If we want to keep this transparent to the users it might make sense to split the namespace into two, one with the functions that are implementation specific and the other with the high-level API.

PR #281, opened 2021-02-04
Generic transactor interface

sundbry:
Hello Datahike Developers!

This adds a pluggable transaction processing engine to the datahike core with minimal interference on the rest of the code base. The transactor is now configurable, with a :transactor key in the main config. If not set, it defaults to a local transaction processor which handles all transactions in a dedicated thread. All existing unit tests pass without modification.

Along with this local implementation, I have developed a proof of concept [transaction processor plugin for Apache Pulsar](https://github.com/arctype-co/datahike-pulsar) (feel free to copy and republish it if you wish to make it an official plugin.) It should be relatively simple to implement something similar using AMQP or Kafka as the centralized transaction log. 

The core idea behind this development is to enable replication of the datahike database in such a way that all instances of a datahike can share a single (distributed) transaction log with a consistent ordering. Thus a transaction applied to any instance will apply to all instances with consistent results. Consistency is only guaranteed to the extent that the backend log service guarantees log message ordering (e.g. running in a Pulsar or Kafka topic with more than one partition may break consistency.)

In the future to address the limitations of a single-partition backend topic, a `:tx-key` option on the `transact` interface could provide a way to guarantee ordering on multi-partition transaction logs. 

This code is quite experimental, your feedback is welcome!

### DIFF
--- a/src/datahike/tools.cljc
+++ b/src/datahike/tools.cljc
@@ -37,14 +37,21 @@
           `(throw #?(:clj  (ex-info (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data)
                      :cljs (error (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data))))))
 
+; (throwable-promise) derived from (promise) in clojure/core.clj.
+; *   Clojure
+; *   Copyright (c) Rich Hickey. All rights reserved.
+; *   The use and distribution terms for this software are covered by the
+; *   Eclipse Public License 1.0 (http://opensource.org/licenses/eclipse-1.0.php)
+; *   which can be found in the file epl-v10.html at the root of this distribution.
+; *   By using this software in any fashion, you are agreeing to be bound by
+; * 	 the terms of this license.
+; *   You must not remove this notice, or any other, from this software.
 (defn throwable-promise
   "Returns a promise object that can be read with deref/@, and set,
   once only, with deliver. Calls to deref/@ prior to delivery will
   block, unless the variant of deref with timeout is used. All
   subsequent derefs will return the same delivered value without
-  blocking. See also - realized?."
-  {:added "1.1"
-   :static true}
+  blocking. Exceptions delivered to the promise will throw on deref."
   []
   (let [d (java.util.concurrent.CountDownLatch. 1)
         v (atom d)]
--- a/src/datahike/tools.cljc
+++ b/src/datahike/tools.cljc
@@ -1,6 +1,7 @@
 (ns ^:no-doc datahike.tools
   (:require
-   [taoensso.timbre :as log]))
+    [superv.async :refer [throw-if-exception-]]
+    [taoensso.timbre :as log]))
 
 (defn combine-hashes [x y]
   #?(:clj  (clojure.lang.Util/hashCombine x y)
@@ -35,3 +36,35 @@
     (list `(log/log! :error :p ~fragments ~{:?line (:line (meta &form))})
           `(throw #?(:clj  (ex-info (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data)
                      :cljs (error (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data))))))
+
+(defn throwable-promise
+  "Returns a promise object that can be read with deref/@, and set,
+  once only, with deliver. Calls to deref/@ prior to delivery will
+  block, unless the variant of deref with timeout is used. All
+  subsequent derefs will return the same delivered value without
+  blocking. See also - realized?."
+  {:added "1.1"
+   :static true}
+  []
+  (let [d (java.util.concurrent.CountDownLatch. 1)
+        v (atom d)]
+    (reify
+      clojure.lang.IDeref
+      (deref [_] (.await d) (throw-if-exception- @v))
+      clojure.lang.IBlockingDeref
+      (deref
+        [_ timeout-ms timeout-val]
+        (if (.await d timeout-ms java.util.concurrent.TimeUnit/MILLISECONDS)
+          (throw-if-exception- @v)
+          timeout-val))
+      clojure.lang.IPending
+      (isRealized [this]
+        (zero? (.getCount d)))
+      clojure.lang.IFn
+      (invoke
+        [this x]
+        (when (and (pos? (.getCount d))
+                   (compare-and-set! v d x))
+          (.countDown d)
+          this)))))
+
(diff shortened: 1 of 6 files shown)


TimoKramer:
Awesome. Will take a look asap.

whilo:
I just want to say this is awesome work! It fits conceptually very well into what we are designing and working on at the moment in terms of distributed Datahike connection management and replication. We might have to generalize it a bit, since we now also have ClojureScript support that will soon be merged into the development branch. @sundbry Are you on our discord server already? Here is an invite link to get there: https://discord.gg/kEBzMvb. It allows us to coordinate more closely with core contributors and the community (WIP).

.run/REPL.run.xml
```diff
@@ -0,0 +1,15 @@
...
+    <setting name="execution" value="DEPS" />
+    <setting name="jvmArgs" value="" />
+    <setting name="parameters" value="" />
+    <setting name="workingDir" value="$PROJECT_DIR$" />
+    <setting name="profiles" value="" />
+    <setting name="aliases" value="dev,test" />
+    <setting name="fixLineNumbers" value="false" />
+    <method v="2" />
+  </configuration>
+</component>
```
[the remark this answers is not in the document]
sundbry: I rebased and removed these configs as unrelated to the core PR, although I still think they are helpful to anyone using IntelliJ/Cursive to hack on this project.

TimoKramer:
I've made some benchmarks and the view is not totally clear.
image

[benchmark.edn.txt](https://github.com/replikativ/datahike/files/6054113/benchmark.edn.txt)
[benchmark-pr.edn.txt](https://github.com/replikativ/datahike/files/6054114/benchmark-pr.edn.txt)

whilo:
Brief update (@sundbry): We will try to merge this work with our migration tool https://github.com/lambdaforge/wanderung/, where we provide an incremental eid mapping (@kordano) between Datomic and Datahike. Hopefully this can be generally done and then it should be able to replicate between a whole set of Datalog databases.

src/datahike/transactor.cljc
```diff
@@ -0,0 +1,54 @@
...
+  (send-transaction! [_ tx-data tx-fn])
+  ; Returns a channel that resolves when the transactor has shut down.
+  (shutdown [_]))
+
+(defrecord LocalTransactor
+  [rx-queue rx-thread]
+  PTransactor
+  (send-transaction! [_ tx-data tx-fn]
+    (let [p (promise-chan)]
+      (>!! rx-queue {:tx-data tx-data :callback p :tx-fn tx-fn})
```
whilo, line 18: I think it would be better to not use a thread-blocking `>!!` here, but rather `put!`. This will return immediately, but we synchronize on the returned promise anyway.

src/datahike/tools.cljc
```diff
@@ -35,3 +36,35 @@
     (list `(log/log! :error :p ~fragments ~{:?line (:line (meta &form))})
           `(throw #?(:clj  (ex-info (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data)
                      :cljs (error (str ~@(map (fn [m#] (if (string? m#) m# (list 'pr-str m#))) msgs)) ~data))))))
+
+(defn throwable-promise
```
whilo, line 40: Adding this is ok, but it would be good to refer here exactly to where it was coming from (and make sure it is under a compatible licence, i.e. EPL v1 compatible). We might replace it with a `promise-chan` later.

src/datahike/transactor.cljc
```diff
@@ -0,0 +1,54 @@
...
+      (loop []
+        (if-let [{:keys [tx-data callback tx-fn]} (<??- rx-queue)]
+          (do
+            (let [update-fn (resolve-fn tx-fn)
+                  tx-report (try (update-and-flush-db connection tx-data update-fn)
+                                 ; Only catch ExceptionInfo here (intentionally rejected transactions).
+                                 ; Any other exceptions should crash the transactor and signal the supervisor.
+                                 (catch clojure.lang.ExceptionInfo e e))]
+              (when (some? callback)
+                (>!! callback tx-report)))
```
whilo, line 39: I would also consider not blocking the transactor on calling back the listening processes, although this was done for Datascript in memory before. Otherwise slow callbacks will jam the transactor. Using `put!` should be fine here, right? This would also render `thread-try` and `<??-` redundant, we can just use a go-loop and run this logic fully asynchronously. As a side-note, the transaction logic here could batch transactions and only flush at the end (i.e. doing updates until the rx-queue is empty and then flush and then notify all sending processes).

whilo: This batching logic could be shared between transactors.

whilo (changes requested):
I thought about it a bit and I think with the few changes to remove blocking core.async calls we should be able to merge.

src/datahike/transactor.cljc
whilo: The only problem that can arise is that there is no direct backpressure on the sender if the queue gets full.

sundbry:
@TimoKramer Interesting benchmark you're right nothing really stands out to me either... more data points / some charts could certainly help. We are dealing with a couple more core.async channels here so that could change the perf slightly. However having all the transactions running on a dedicated thread might also make cpu caching better / total thruput higher. IDK for sure! This reminds me of a project ["Gorilla REPL"](http://gorilla-repl.org/) which might come in handy here for analyzing the benchmarks.

src/datahike/transactor.cljc
sundbry: `>!!` won't block if the buffer has space (promise chan has 1 slot so it wont be full because nothing else is putting onto it) however I will change it to use `put!` here anyways since >!! uses put! under the hood so it's just less code to execute.

sundbry: > This would also render thread-try and <??- redundant, we can just use a go-loop and run this logic fully asynchronously.

Can't do that because the loop is io bound on the update-and-flush. You're not supposed to do significant IO on the go thread pool. This also keeps the CPU cache "hot" on the main loop here if it matters at all (it might under real stress) - less context switching.

> batch transactions and only flush at the end (i.e. doing updates until the rx-queue is empty and then flush and then notify all sending processes).

I'm sure something like this is possible but also needs to be carefully implemented. Like a max timeout/# of transactions before a flush is forced. I would save this for a follow up PR as a performance optimization. We don't want callbacks to be blocked for a long time if the queue is long for example, because that could lead to a breakdown in performance under load. There is probably a 'sweet spot' to hit on these parameters.

src/datahike/transactor.cljc
sundbry: @whilo Yes the back pressure is exactly why I intentionally used >!! here.
If we provide a method using `put!` I think it should come with an alternative public interface which lets the client know when the put is not accepted if the buffer is full.

whilo:
Looks good now, thanks! Can you do a rebase? I would also let @kordano or @grischoun review it before merging.

whilo (approved):

grischoun (approved):
Looks very good to me.

grischoun:
Closing this PR as it is replaced by https://github.com/replikativ/datahike/pull/310 which fixes the current conflicts.

PR #282, opened 2021-02-08
Refactor integration tests

TimoKramer:
With this change a new namespace is introduced that can be used
    to implement integration-tests from all subprojects. This enables
    us to run integration tests everywhere without copying the tests
    around. Like in konserve.

### DIFF
--- a/src/datahike/integration_test.cljc
+++ b/src/datahike/integration_test.cljc
@@ -1,0 +1,65 @@
+(ns datahike.integration-test
+  (:require [datahike.api :as d]
+            #?(:clj [clojure.test :refer :all]
+               :cljs [cljs.test :refer :all :include-macros true])))
+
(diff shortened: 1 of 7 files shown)


dev/sandbox.clj
```diff
@@ -1,5 +1,7 @@
 (ns sandbox
-  (:require [datahike.api :as d]))
+  (:require [datahike.api :as d]
+            [datahike-leveldb.core]
+            [datahike-postgres.core :as pg]))
```
kordano, line 4: Are both `backends` always needed here if we remove them in the future?

TimoKramer: sorry, I guess the sandbox was a mistake.

kordano: Great, thanks for the refactoring.

kordano (approved):

PR #288, opened 2021-02-23
Add benchmark to CI

TimoKramer:


TimoKramer:
Do you think we should have a more elaborate tagging mechanism. I would like to save the running env like vcpus and mem for each run.

.circleci/config.yml
```diff
@@ -40,6 +40,18 @@ jobs:
...
+      image: ubuntu-2004:202101-01
+    environment:
+      TIMBRE_LEVEL: ":warn"
+    steps:
+      - attach_workspace:
+          at: /home/circleci
+      - run: curl -O https://download.clojure.org/install/linux-install-1.10.2.796.sh
+      - run: chmod +x linux-install-1.10.2.796.sh
+      - run: sudo ./linux-install-1.10.2.796.sh
+      - run: cd ~/replikativ && clojure -M:benchmark --db-server-url ${DB_URL} --db-name ${DB_NAME} --db-token ${DB_TOKEN} --tag ${CIRCLE_SHA1}
```
jsmassa, line 54: What kind of information is in  CRIRCLE_SHA1?

jsmassa: Mind you that so far we can already add as many tags as we want to. The tagging system is designed to be very general, so it can hold any type of information we think is important. What I had in mind when writing it was information to filter the data in a UI, so especially I thought about information to identify the branch or even commits.

If we can access the system information though, we can also add that, that would make sense. The question is if we want to use the tagging then or introduce separate attributes for that.

TimoKramer: My idea was to use the tagging for generic attributing. I don't yet know which data we might need in the future but there might be more and if we just use k/v-pairs for tagging we could make it something like this:
`vcpus: 2`
`mem: 4gb`
`gitsha: 1a3b4d`
`weather: sunny`
etc...

TimoKramer: > What kind of information is in CRIRCLE_SHA1?

that's the git commit sha

jsmassa: I guess we are thinking of something similar or the same then. My idea was also that e.g. for the branch name, we would add "--tag branch:branch-name"  and for your example that would be
"clj -M:benchmark --tag vcpus:2 --tag mem:4gb --tag gitsha:1a3b4d --tag weather:sunny".
Is that also what you were thinking of?

TimoKramer: right, is this already the way it works? sorry, I could have probably looked it up easily:)

whilo:
I just want to say that I think integrating this is awesome :100: .

.circleci/config.yml
TimoKramer: ok, so you think this is sufficient? We just save the tags as `str:str` and then we just deserialize on fetching the data? Fine for me, was just wondering. Then I think we can accept this PR as is, if you don't have any other concerns. Thanks @jsmassa

jsmassa: For now, I think it is sufficient. We still can change it later if we see that it is not working for us that way.

jsmassa (approved):

PR #291, opened 2021-03-04
Add standard deviations to benchmarks

jsmassa:


kordano:
Is this good to go?

whilo:
I would say so. I am not sure what `criterium` exactly does, but I remember it calculating median and quantiles. This is better because they are robust towards outliers and more informative, but I think we can fix that also later.

kordano (approved):

PR #292, opened 2021-03-25
289 upsert fails on repeated value

grischoun:
Fixes the issue by adding a timestamp to upsert operations.

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -1251,7 +1251,7 @@
         true (advance-max-eid (.-e datom))
         true (update :hash + (hash datom))
         schema? (-> (update-schema datom)
-                  update-rschema)
+                    update-rschema)
         true (update :op-count + 1))
       (if-some [removing ^Datom (first (-search db [(.-e datom) (.-a datom) (.-v datom)]))]
         (cond-> db
@@ -1334,7 +1334,7 @@
   (let [indexing?     (indexing? db (.-a datom))
         schema?       (ds/schema-attr? (.-a datom))
         keep-history? (and (-keep-history? db) (not (no-history? db (.-a datom)))
-                        (not= :db/txInstant (.-a datom)))
+                           (not= :db/txInstant (.-a datom)))
         op-count      (.op-count db)]
     (cond-> db
       ;; Optimistic removal of the schema entry (because we don't know whether it is already present or not)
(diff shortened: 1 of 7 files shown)


src/datahike/db.cljc
```diff
@@ -1237,30 +1241,33 @@
...
+          true (update-in [:aevt] #(di/-remove % removing :aevt op-count))
+          indexing? (update-in [:avet] #(di/-remove % removing :avet op-count))
           true (update :hash - (hash removing))
           schema? (-> (remove-schema datom) update-rschema)
-          keep-history? (update-in [:temporal-eavt] #(di/-insert % removing :eavt))
-          keep-history? (update-in [:temporal-eavt] #(di/-insert % datom :eavt))
-          keep-history? (update-in [:temporal-aevt] #(di/-insert % removing :aevt))
-          keep-history? (update-in [:temporal-aevt] #(di/-insert % datom :aevt))
+          keep-history? (update-in [:temporal-eavt] #(di/-insert % removing :eavt op-count))
+          keep-history? (update-in [:temporal-eavt] #(di/-insert % datom :eavt (+ 1 op-count)))
```
whilo, line 1264: Minor detail, it is more idiomatic in Clojure to use `inc` instead of `+ 1`.

src/datahike/index/hitchhiker_tree/upsert.cljc
```diff
@@ -43,8 +44,9 @@
       ;; '-' means it is retracted and 'nt' is the current transaction time.
       [a b c (- nt)])))
 
-(defrecord temporal-UpsertOp [key]
+(defrecord temporal-UpsertOp [key ts]
```
whilo, line 47: I think it would have been better to name this `TemporalUpsertOp`. We can rename it, you just need to add a read-handler that automatically transforms the old records to the new name. But we can also do so later.

whilo (changes requested):
Nice work. For some reason the dependencies for konserve are currently broken, this might be a problem with bumping the hitchhiker-tree release, I can fix this. I think it would be good to load max-tx value as initial op-count for existing databases as discussed, then I can automatically extract those when reading InsertOps and DeleteOps without timestamps in the hh-tree repository.

grischoun:
@whilo @kordano All tests pass. 
It is currently pointing to the hitchhiker-tree branch that it needs as a dependency.

whilo (approved):

PR #306, opened 2021-04-15
Adds the hash-join optimization.

grischoun:
The speedup comes from building the hash-table on the smallest of the two tuple sets.

### DIFF
--- a/src/datahike/query.cljc
+++ b/src/datahike/query.cljc
@@ -417,32 +417,46 @@
       (persistent! hash-table))))
 
 (defn hash-join [rel1 rel2]
-  (let [tuples1 (:tuples rel1)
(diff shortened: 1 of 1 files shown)


whilo (approved):

jsmassa (commented):
So this edit simply adds the decision if tuple set 1 will be merged in tuple set 2 or tuple set 2 will be merged in tuple set 1? 

So I think we should only see differences in performance with this when at some point the two sets for have significantly different sizes. Especially with the following query I implemented, I would expect a change then: 
`{:function :one-join-query-second-fixed
      :query (one-join-query-second-fixed db :i1 :i2 (rand-nth known-i2))
      :details {:data-type :int :data-in-db? true}}` 
How are the results for this one?

PR #315, opened 2021-05-05
WIP: Add caching

jsmassa:
Add caching to index search and other functions

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -6,7 +6,7 @@
    #?(:clj [clojure.pprint :as pp])
    [datahike.index :refer [-slice -seq -count -all -persistent! -transient] :as di]
    [datahike.datom :as dd :refer [datom datom-tx datom-added datom?]]
-   [datahike.constants :refer [e0 tx0 emax txmax]]
+   [datahike.constants :refer [e0 tx0 emax txmax db-cache-size]]
    [datahike.tools :refer [get-time case-tree raise]]
    [datahike.schema :as ds]
    [datahike.lru :as lru]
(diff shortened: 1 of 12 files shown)


jsmassa:
Only implements caching for search at the moment. Maybe it should be implemented on the index -slice function instead though since that is one of the worst bottlenecks. 
Another thing we have to think about before merging is how when the cache should be cleared. Maybe we have to implement a garbage collector for that.

jsmassa:
A trivial version of clearing caches is now implemented

kordano:
Could you convert to Draft if this is still a WIP?

kordano:
@whilo what are your thoughts on caching?

src/datahike/db.cljc
```diff
@@ -677,6 +656,58 @@
 
 ;; ----------------------------------------------------------------------------
 
+(def db-caches (atom {}))
+
+(defn insert-cache [db-hash cache]
+  (let [caches @db-caches]
+    (if (>= (count caches) max-db-caches)
+      (let [[last-used-db _] (reduce (fn [[d-hash1 [c1 ^Date last-accessed1]]
```
whilo, line 664: It would be better to use a LRU cache here, we use `clojure.core.cache` everywhere else, have you taken a look at that? Your reduce takes linear time in the size of the cache.

whilo (changes requested):
In principle that looks mergeable, but I think it would be good to use `clojure.core.cache` if possible. We can also discuss details in our next meeting.

src/datahike/db.cljc
jsmassa: Hadn't heard of that `clojure.core.cache` yet. Will have a look if at if is usable in this case

PR #316, opened 2021-05-05
WIP: Add type hints

jsmassa:
Version with all reflection warnings from datahike resolved and more type hints to resolve [unboxed math warnings](https://insideclojure.org/2014/12/15/warn-on-boxed/).

### DIFF
--- a/benchmark/src/benchmark/compare.clj
+++ b/benchmark/src/benchmark/compare.clj
@@ -49,14 +49,12 @@
                           (map #(assoc % :source filename)
                                (edn/read-string (slurp filename))))
                         filenames)
-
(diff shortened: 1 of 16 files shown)


jsmassa:
So far it looks like it makes the performance rather worse than better. My guess is added complexity due to type casts.
Still have to check the performance though in the bigger context.

kordano:
Please convert to draft PR if it is not ready yet.

jsmassa:
It is basically ready. We just should look at the benchmark results together to decide if we want to include it

jsmassa:
Closing due to to unconvincing benchmark results.

Issue #308, opened 2021-04-21
Pull with reverse attribute returns duplicate results


rafd:
After transacting data multiple time, a pull query with a reverse attribute includes a copy of result for *each* time it was transacted.

For example: 

If I transact the following:
```clojure
[{:event/id 1}
 {:talk/id 10
  :talk/event [:event/id 1]}
 {:talk/id 20
  :talk/event [:event/id 1]}])
```
When I pull using:
```clojure
[:event/id
 {:talk/_event [:talk/id]}]
```
I expect to get the following (no matter the number of times I transact the above):
```clojure
{:event/id 1
 :talk/_event [{:talk/id 10}
               {:talk/id 20}]}
```
But, if I transact 2 times, I get this instead:
```clojure
{:event/id 1
 :talk/_event [{:talk/id 10}
               {:talk/id 10}
               {:talk/id 20}
               {:talk/id 20}]}
```
Note the duplicate objects in the `:talk/_event` value.

If I transact 3 times, there would be 3 of each; 4 times, 4 each, etc.

When querying without pull, the database returns the correct number of events (1) and talks (2).

Here is a test file: https://gist.github.com/rafd/0262604c49b4cb2b9544ae33e35e396e

This issue does not exist in datascript (likely because datascript does not keep history).

rafd:
Upon further investigation, the underlying cause is that the `:avet` index is erroneously returning all historical datoms, rather than those at current `t`.

A slightly modified version of `datahike.test.index/test-datoms`, where facts are transacted twice, fails the `:avet` tests:

```clojure
(deftest test-datoms
  (let [dvec #(vector (:e %) (:a %) (:v %))
        txs [[:db/add 1 :name "Petr"]
             [:db/add 1 :age 44]
             [:db/add 2 :name "Ivan"]
             [:db/add 2 :age 25]
             [:db/add 3 :name "Sergey"]
             [:db/add 3 :age 11]]
        db (-> (d/empty-db {:age {:db/index true}})
               (d/db-with txs) ;; change: repeated twice
               (d/db-with txs))]
    (testing "Main indexes, sort order"
      (is (= [[1 :age 44]
              [2 :age 25]
              [3 :age 11]
              [1 :name "Petr"]
              [2 :name "Ivan"]
              [3 :name "Sergey"]]
             (map dvec (d/datoms db :aevt))))

      (is (= [[1 :age 44]
              [1 :name "Petr"]
              [2 :age 25]
              [2 :name "Ivan"]
[...]

PR #320, opened 2021-05-06
308 fix avet upsert

grischoun:
Fixes issue with upserting on :avet.

### DIFF
--- a/src/datahike/index/hitchhiker_tree.cljc
+++ b/src/datahike/index/hitchhiker_tree.cljc
@@ -125,14 +125,20 @@
 (defn -insert [tree ^Datom datom index-type op-count]
   (hmsg/insert tree (datom->node datom index-type) nil op-count))
 
+(defn- index-type->indices [index-type]
+  (case index-type
+    :eavt [0 1]
+    :aevt [0 1]
+    :avet [0 2]
+    (throw (UnsupportedOperationException. "Unknown index type: " index-type))))
 (defn -upsert [tree ^Datom datom index-type op-count]
   (let [datom-as-vec (datom->node datom index-type)]
-    (async/<?? (hmsg/enqueue tree [(assoc (ups/new-UpsertOp datom-as-vec op-count)
+    (async/<?? (hmsg/enqueue tree [(assoc (ups/new-UpsertOp datom-as-vec op-count (index-type->indices index-type))
                                           :tag (h/uuid))]))))
 
 (defn -temporal-upsert [tree ^Datom datom index-type op-count]
   (let [datom-as-vec (datom->node datom index-type)]
-    (async/<?? (hmsg/enqueue tree [(assoc (ups/new-temporal-UpsertOp datom-as-vec op-count)
+    (async/<?? (hmsg/enqueue tree [(assoc (ups/new-temporal-UpsertOp datom-as-vec op-count (index-type->indices index-type))
                                           :tag (h/uuid))]))))
 
 (defn init-tree
--- a/src/datahike/index/hitchhiker_tree.cljc
+++ b/src/datahike/index/hitchhiker_tree.cljc
@@ -131,6 +131,7 @@
     :aevt [0 1]
     :avet [0 2]
     (throw (UnsupportedOperationException. "Unknown index type: " index-type))))
+
 (defn -upsert [tree ^Datom datom index-type op-count]
   (let [datom-as-vec (datom->node datom index-type)]
     (async/<?? (hmsg/enqueue tree [(assoc (ups/new-UpsertOp datom-as-vec op-count (index-type->indices index-type))
--- a/src/datahike/index/hitchhiker_tree/upsert.cljc
+++ b/src/datahike/index/hitchhiker_tree/upsert.cljc
@@ -2,6 +2,13 @@
   (:require [hitchhiker.tree :as tree]
             [hitchhiker.tree.op :as op]))
 
+
+(defn- max-t
+  "Returns the key with the max 't' component"
+  [ks]
+  (when (seq ks)
+    (apply max-key #(nth % 3) ks)))
+
 (defn old-key
   "Returns the old version of the given 'new' key if it exists in 'old-keys'.
   If there are multiple old versions, the one with the biggest transaction time is returned.
@@ -20,9 +27,7 @@
                                 (and bool (= (nth % i) (nth new i))))
                               true
                               indices))
-             ((fn [ks]
-                (when (seq ks)
-                  (apply max-key #(nth % 3) ks)))))))))
+             max-t)))))
 
 (defn remove-old
   "Removes old key from the 'kvs' map using 'remove-fn' function if 'new' and 'old' keys' first two entries match."
--- a/src/datahike/index/hitchhiker_tree/upsert.cljc
+++ b/src/datahike/index/hitchhiker_tree/upsert.cljc
@@ -2,8 +2,21 @@
   (:require [hitchhiker.tree :as tree]
             [hitchhiker.tree.op :as op]))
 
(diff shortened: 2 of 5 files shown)


grischoun:
This fixes issue #308.

grischoun:
I ran the benchmarks against this change and there is no change in performance compared to what we had before (current dev. branch).

jsmassa (dismissed):
Nice work

whilo:
I need to look into the details still, but have you made sure that the old `UpsertOp` values can still be read correctly by the read handlers? I guess we still don't have automatic upgrade tests.

grischoun:
> I need to look into the details still, but have you made sure that the old `UpsertOp` values can still be read correctly by the read handlers? I guess we still don't have automatic upgrade tests.

Yes, read handlers are working. 
Old UpsertOps could end up with a nil value for the newly introduced field `indices` but this line protects against that: https://github.com/replikativ/datahike/blob/07b0ea4bd5576f13bd6bd08feff3ff4ad030fc3e/src/datahike/index/hitchhiker_tree/upsert.cljc#L10
This leads to the safe but slower path.
I think it is the best we can do for now. May be later with Wanderung, knowing on which kind of tree we are working on, we will be able to fix it by setting the correct value for `indices`.

whilo:
Is there a default value that can be used on the old values being read? It is faster and cleaner to extend the read handler than to substitute values in the logic.

src/datahike/index/hitchhiker_tree/upsert.cljc
```diff
@@ -2,54 +2,86 @@
...
+  [old-keys new indices]
+  (when (seq old-keys)
+    (let [mask (mask new indices)]
+      (when-let [candidates (subseq old-keys >= mask)]
         (->> candidates
              (map first)
-             (take-while #(and (= a (first %)) (= b (second %))))
-             reverse
-             first)))))
+             ((if (prefix? indices) take-while filter)
```
grischoun, line 41: What is going on here?
When the 'indices' in the vector are strictly increasing by 1, like in [0 1 2], then we know that we are dealing with results returned from 'subseq' which are sorted from the 1st, 2nd and third entries in the datom. Thus we can use 'take-while' to retrieve all matching datoms until there is no match.
But when the 'indices' vector is like [0 2], which is the case for :avet, 'subseq' won't return datoms sorted with 'a' then 'e' but rather with 'a' then 'v' and 'e'. Thus we can't use 'take-while' but are forced to filter out all the results from 'subseq' that match 'a' and 'e'.
In summary, 'prefix?' is used to check whether the 'indices' vector represents an index where the prefix of datoms are used to sort them out or not.

whilo: Ok, thank you for explaining!

whilo: Ok. Thank you for explaining!

src/datahike/index/hitchhiker_tree/upsert.cljc
```diff
@@ -2,54 +2,86 @@
...
 (defn old-retracted
   "Returns a new datom to insert in the tree to signal the retraction of the old datom."
-  [kvs key]
-  (when-let [old (old-key kvs key)]
+  [kvs key indices]
+  (when-let [old (old-key kvs key indices)]
     (let [[a b c _] old
           [_ _ _ nt] key]
       ;; '-' means it is retracted and 'nt' is the current transaction time.
       [a b c (- nt)])))
```
whilo, line 77: Do we use negative transaction times for retractions in general?

whilo (approved):
Looks good to me.

src/datahike/index/hitchhiker_tree.cljc
```diff
@@ -119,14 +119,21 @@
 (defn -insert [tree ^Datom datom index-type op-count]
   (hmsg/insert tree (datom->node datom index-type) nil op-count))
 
+(defn- index-type->indices [index-type]
+  (case index-type
+    :eavt [0 1]
+    :aevt [0 1]
+    :avet [0 2]
```
whilo, line 126: Is this true if we operate on the `avet` index itself? This is only true if we operate on `aevt`, right?

whilo (changes requested):
Maybe you can clarify this before merging.

src/datahike/index/hitchhiker_tree/upsert.cljc
grischoun: That's what I always thought. @kordano Can you confirm this?

src/datahike/index/hitchhiker_tree.cljc
grischoun: @whilo Sorry, I don't understand the question.

whilo: I meant the only case with `[0 2]`, `:avet`. I guess we do not work on the actual `avet` index in this case, otherwise it would be `[0 1]`.

grischoun: Yes, we don’t work on avet.

whilo: Ok, can you explain why not? I would have guessed that we need to upsert in there as well(?). Sorry if this is a question with an obvious answer.

grischoun: Sorry my previous answer is based on a misunderstanding of your question. 
We *do* upsert on :avet but we need to use :a and :e, i.e., position 0 and 2 (and not 0 and 1) to correctly perform upserts on :avet.

grischoun:
This PR also fixes #338 .

src/datahike/index/hitchhiker_tree/upsert.cljc
```diff
@@ -58,28 +90,28 @@
...
-          (fn [{:keys [key value]}]
-            (map->UpsertOp {:key key :value value}))
+          (fn [{:keys [key value ts indices]}]
+            (map->UpsertOp {:key key :value value :ts ts :indices (or indices [0 1])}))
 
           'datahike.index.hitchhiker_tree.upsert.temporal-UpsertOp
-          (fn [{:keys [key value]}]
-            (map->temporal-UpsertOp {:key key :value value}))})
+          (fn [{:keys [key value ts indices]}]
+            (map->temporal-UpsertOp {:key key :value value :ts ts :indices (or indices [0 1])}))})
```
grischoun, line 116: These are the fixes for #338.

src/datahike/index/hitchhiker_tree.cljc
whilo: Thanks, that makes sense.

whilo (approved):

whilo (changes requested):
I think it is important to either go with `op-count` everywhere or `ts`, but not to have both. As far as I see here this is still the case for `UpsertOp` and `temporal-UpsertOp`. I think you said that you are still fixing this, I am just making sure that my review is uptodate and we can merge as soon as it is ready.

grischoun:
> I think it is important to either go with `op-count` everywhere or `ts`, but not to have both. As far as I see here this is still the case for `UpsertOp` and `temporal-UpsertOp`. I think you said that you are still fixing this, I am just making sure that my review is uptodate and we can merge as soon as it is ready.


I don't think we can rewrite everything to `op-count` (or `ts`) for now because there might be existing databases that have been serialised with `ts`. Once we have 'Wandering' though, we could migrate these databases and get rid of `ts`.

grischoun:
'ts' is replaced by 'op-count' now. 
Ready for another round of reviews.

whilo (approved):
This looks good to me.

PR #323, opened 2021-05-07
319 cleanup db records

kordano:


doc/adl/adr-002-db-record.md
```diff
@@ -0,0 +1,63 @@
...
+
+![datahike-db-record](../assets/datahike-db-record.svg)
+
+Additionally simplify `DB` copy process after connections and transactions.
+
+#### Pro
+
+- unified and cleaner DB record
+- more extensible through nested structures
+- creates prerequisits for future metrics and additional meta data like migration history that needs to be stored in the record
```
grischoun, line 29: prerequisites

doc/adl/adr-002-db-record.md
```diff
@@ -0,0 +1,63 @@
...
+
+- possibily breaking changes if not every old field is mapped to the new implementation
+- migration path adds more friction
+- a lot of documentation needs to be adjusted
+- refactoring needs to be done on sensitive code (`datahike.api`, `datahike.connector`, `datahike.db`)
+
+### Extension
+
+Instead of refactoring all new fields should be added into categories as proposed in the first [option](#refactoring).
+
```
grischoun, line 41: The difference between 'Refactoring' and 'Extension' is not clear to me.
What would the data structure look like in the 'Refactoring' case?

grischoun (commented):
Overall, the idea makes a lot of sense.

doc/adl/adr-002-db-record.md
```diff
@@ -0,0 +1,62 @@
...
+Each change requires updating the `IConfiguration` from `datahike.connector` where the store is updated and a new `DB` record is written.
+
+## Options
+
+### Refactoring
+
+Adjust the database record `DB` and combine fields into categories. Proposed fields:
+
+- `indexes`: all index nodes
+- `config`: as it is now with cleanup
```
TimoKramer, line 15: I don't understand 'as it is now with cleanup'

TimoKramer (commented):
Good idea to refactor this although I did not have pain with this so far.

whilo:
I think this is a good idea. An important consistency check is that this itself defines a schema over internal representations of Datahike that from now on should be transparent in the RDF sense. That means any reader of a DB record needs to be able to determine which schema of this representation it is using.

PR #324, opened 2021-05-07
String comparator fix

jsmassa:
This PR enables queries with `<` and `>` predicates for strings. `=` seems to be working already. There might be a better way to solve this, but at least it works now.

### DIFF
--- a/src/datahike/query.cljc
+++ b/src/datahike/query.cljc
@@ -204,10 +204,17 @@
 
 (defmulti -lesser?
   {:arglists '([value & more])}
-  (fn [value & more] (class value)))
+  (fn [value & more] 
+    (class value)))
 
-(defmethod -lesser? String [^String s0 ^String s1]
-  (neg? (compare s0 s1)))
+(defmethod -lesser? java.lang.String [^String s0 & more]
+  (reduce (fn [res [s1 ^String s2]]
+            (if (neg? (compare s1 s2))
+              res
+              (reduced false)))
+          true
+          (partition 2 1 (cons s0 more))))
+
 
 (defmethod -lesser? java.util.Date [^Date d0 ^Date d1]
   #?(:clj  (.before ^Date d0 ^Date d1)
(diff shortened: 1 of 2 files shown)


src/datahike/query.cljc
```diff
@@ -204,7 +204,15 @@
...
   {:arglists '([value & more])}
-  (fn [value & more] (class value)))
+  (fn [value & _more] (class value)))
+
+(defmethod -lesser? java.lang.String [^String s0 & more]
+  (reduce (fn [res [s1 ^String s2]]
+            (if (neg? (compare s1 s2))
+              res
+              (reduced false)))
+          true
```
whilo, line 214: Is it sound to have both `-lesser?` and `-greater?` default to `true`?

whilo (commented):
What is the original motivation for this fix? There is no issue assigned to the PR, which would be better to provide context, especially when it is a fix.

src/datahike/query.cljc
jsmassa: It's the same with `<` and `>`, so I would say: yes

jsmassa:
I found this problem during benchmarking for strings. I needed it fixed, so I fixed it. To replicate it, try the tests I wrote without the fix. We might want to write tests for the other implemented datatypes as well. The date comparator for example looks suspicious to me as well

src/datahike/query.cljc
whilo: I think this is mathematically questionable, because `<` and `>` are strict binary relations which cannot both be true for the same arguments, but if this is expected behaviour in Clojure then this is what it is. It would be better to always require at least 2 arguments to assign meaning to `<` or `>`. This is different for `+` or `*` defaulting to the neutral elements in their group for no arguments, which I think is mathematically reasonable.

whilo (approved):

whilo:
@jsmassa Is it possible for us to use protocols instead of multimethods for the comparators of `-lesser?` and `-greater?`? Multimethods are comparatively slow and we seem to dispatch on the type of the first argument here. This is exactly what OO dispatch is made for and is reflected in Clojure's protocols. Here is a bit of background https://stackoverflow.com/questions/8070368/clojure-multimethods-vs-protocols.

jsmassa:
It should be possible and I think it would also be cleaner. We should then implement the protocols for all of our datatypes then we can also be sure that it works. Additionally to lesser, greater and equals, I would also implement min and max since I will need that for the predicate optimization. 
I would suggest to rename issue #328 though and tackle the changes there

PR #332, opened 2021-05-21
feat: distributed reads and transactor for datahike-server

### LINKED ISSUE #322: 1 writer, n reader
I cannot find a forum to ask questions for, so I put the question in here:

Is it safe to have multiple connections with one writer app and n reader apps connected to the
same file storage? Will the readers pick up new data?

whilo:
This PR implements distributed read operations by redirecting the access to the connection to the underlying konserve storage to update the DB value (closes #322). It requires one roundtrip to fetch the root under `:db` and is a polling mechanism that is efficient if the reader is on average not accessing the database more often than it is being written to it and if the latency of the store access is not too high for the application. For low latency the transactor interface now exposes whether it is actively streaming, as is the default local transactor that we had so far (basically directly operating on the connection atom). 

This PR also provides an implementation of the transactor interface for datahike-server, which provides the single writer to the underlying store and is not able to stream yet. Provided datahike server is running **with the same datahike version** and configured with the example configuration, the following configuration works now for an arbitrary number of Datahike peers (readers):

```clojure
{:store    {:backend :file :path "/tmp/dh-file"}
            :keep-history? true ;; true
            :schema-flexibility :read
            :transactor {:backend :datahike-server
                         :client-config  {:timeout 300
                                          :endpoint "http://localhost:3333"
                                        ;:token "securerandompassword"
                                          :db-name "users"}}
            }
```
It is not clear to me yet whether we want to keep this transactor in this repository or move it into a separate one because we now have an additional dependency on `datahike-client` that we can avoid. Probably it is better to pull it out. The benefit of including it in Datahike is that Datahike will then always work with datahike-server instances out of the box.


TimoKramer:
datahike-client will not work either with datahike when compiled with JDK<11 because I am using the JDK built-in http-client in the datahike-client

deps.edn
```diff
@@ -8,7 +8,9 @@
         io.replikativ/superv.async                  {:mvn/version "0.2.11"}
         io.lambdaforge/datalog-parser               {:mvn/version "0.1.8"}
         io.replikativ/zufall                        {:mvn/version "0.1.0"}
-        junit/junit                                 {:mvn/version "4.13.1"}}
+        junit/junit                                 {:mvn/version "4.13.1"}
+        io.replikativ/datahike-client               {:git/url "https://github.com/replikativ/datahike-client.git"
```
TimoKramer, line 12: Just made a 0.1.0-SNAPSHOT

TimoKramer (commented):
Awesome! Do you intend to add the streaming within this PR?

alekcz:
> datahike-client will not work either with datahike when compiled with JDK<11 because I am using the JDK built-in http-client in the datahike-client

@TimoKramer I've submitted a [pull req](https://github.com/replikativ/datahike-client/pull/1) to make the client work on JDK8

whilo:
@alekcz Does http-kit also work with Graal's native-images? We are currently heading in this direction and I think it will be particularly beneficial if our HTTP client is also embeddable into native images.

TimoKramer:
babashka supports http-kit so this seems to work. I think it is a good choice because with borkdude supporting it it is obviously very well alive:)

whilo:
@TimoKramer I think Datahike should be usable without depending on libraries for network IO. Therefore I think it would not be bad to move the code of the actual `DatahikeServerTransactor` and with it the dependencies on a HTTP client out of this PR into a separate downstream repository that can be included into projects as needed. I would do the same for the native-image cli interface. Nonetheless it could potentially include the `DatahikeServerTransactor`.

alekcz:
> @alekcz Does http-kit also work with Graal's native-images? We are currently heading in this direction and I think it will be particularly beneficial if our HTTP client is also embeddable into native images.

Yes it does. It's actually a built-in library in babashka. now too.

whilo:
@jsmassa @kordano This PR is basically complete and ready for feedback. I still need to test it with datahike-server and its transactor and the tests turned out to be actually broken with proper connection management, so they also still need to be fixed. In particular many tests use the same storage configuration and then trip over each other when they are run in parallel by kaocha. 

@TimoKramer something is also a bit weird with this branch here, I had some difficulties force pushing it and now it shows me conflicts here that should not be there. If you happen to spot the problem, please let me know.

whilo:
This PR also covers https://github.com/replikativ/datahike/pull/439 btw. You can now pass in `tx-meta`a  `noCommit true` flag to avoid writing to disk in a transaction. The benefit of this is that it is tracked as data per transaction instead of transactor state that could interleave with other transactions being dispatched or still pending.

test/datahike/test/core_test.cljc
```diff
@@ -116,60 +116,3 @@
           (dc/db-with (dc/empty-db) [{:a 1 :b 2 :c 4} {:a 1}])
           (dc/db-with (dc/empty-db) [{:a 1 :b 3 :d 5}])))))
 
-;; connection tests
```
jsmassa, line 119: Why are these tests gone now?

test/datahike/test/migrate_test.clj
```diff
@@ -117,9 +117,9 @@
         (let [cfg (merge base-config
                          {:keep-history? hist
                           :attribute-refs? attr-ref})
-              old-cfg (assoc-in cfg [:store :path] old-path)
+              old-cfg (assoc-in cfg [:store :path] (str old-path (.getTime (java.util.Date.))))
```
jsmassa, line 120: It seems to me like it would make sense to have a little helper function that creates ids based on time. This would make it more explicit what this string is there for as the usage of date objects makes it a bit confusing.

I would also suggest to always create a function in `datahike.tools` or even in a special compatibility namespace when java functions are used in order to facilitate the upkeep of clojurescript compliance. For the creation of a date object, there is already a function currently named `get-time` in `datahike.tools` that works with clj and cljs.

whilo: Yes, I built konserve, kabel, replikativ etc. that way, where we have code in a `platform` namespace for each platform. This is fixed now.

src/datahike/writing.cljc
```diff
@@ -0,0 +1,238 @@
...
+     (reset! connection db)
+     (if noCommit
+       tx-report
+       (assoc-in tx-report [:tx-meta :db/commitId]
+                 (get-in db [:meta :datahike/commit-id]))))))
+
+(defprotocol PDatabaseCreator
+  (-create-database [config opts])
+  (-delete-database [config])
+  (-database-exists? [config]))
```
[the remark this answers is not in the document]
whilo: I changed it to `PDatabaseManagement`, which is more fitting. This functionality is happening on a meta level of normal DB operations.

test/datahike/test/core_test.cljc
whilo: Because you cannot instantiate connections directly anymore, you need to go through the connector now (which deduplicates and ensures that there is always only one connection to a store.

src/datahike/store.cljc
```diff
@@ -107,12 +120,16 @@
 
 (defmethod default-config :file [config]
   (merge
-   {:path (:datahike-store-path env "datahike-db")}
+   {:path (:datahike-store-path env "datahike-db")
+    :scope #?(:clj (.getHostAddress (InetAddress/getLocalHost))
+              :cljs (throw (ex-info "Not supported yet." {})))}
```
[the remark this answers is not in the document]
whilo: Fixed. There are other exceptions thrown in this namespace manually though, which is probably because they are `IllegalArgumentException`s, which is not possible to do with `dt/raise`.

src/datahike/config.cljc
```diff
@@ -185,6 +192,10 @@
        (update merged-config :initial-tx (fn [path] (-> path slurp read-string)))
        merged-config))))
 
+(defn config-merge [stored-config config]
```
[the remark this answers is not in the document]
whilo: Yes, I am using clj-kondo and was advocating for everyone in LF to use it when it came out. Unfortunately my version did not lint it, I might have to update it.

src/datahike/api.cljc
```diff
@@ -178,17 +194,20 @@
...
-  transact!
-  dc/transact!)
+  (fn [connection arg-map]
+    (let [arg (cond
+                (and (map? arg-map) (contains? arg-map :tx-data)) arg-map
+                (vector? arg-map)                                 {:tx-data arg-map}
+                (seq? arg-map)                                    {:tx-data arg-map}
+                :else                                             (dt/raise "Bad argument to transact, expected map with :tx-data as key.
+                               Vector and sequence are allowed as argument but deprecated."
+                                                                            {:error :transact/syntax :argument arg-map}))]
```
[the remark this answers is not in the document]
whilo: I just moved this code there from the old `connection.cljc`. I don't fully understand what you mean. Feel free to directly patch it or describe what I should change.

test/datahike/test/core_test.cljc
whilo: These tests were also not really providing any important coverage as far as I could see.

jsmassa: I see, I believe you are right about the coverage.

src/datahike/api.cljc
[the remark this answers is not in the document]
jsmassa: I changed it accordingly.

src/datahike/api.cljc
```diff
@@ -196,12 +197,18 @@
...
   (fn [connection arg-map]
     (let [arg (cond
-                (and (map? arg-map) (contains? arg-map :tx-data)) arg-map
-                (vector? arg-map)                                 {:tx-data arg-map}
-                (seq? arg-map)                                    {:tx-data arg-map}
-                :else                                             (dt/raise "Bad argument to transact, expected map with :tx-data as key.
-                               Vector and sequence are allowed as argument but deprecated."
-                                                                            {:error :transact/syntax :argument arg-map}))]
+                (or (vector? arg-map)
+                    (seq? arg-map))   (do (log/warn "Vector and sequence as argument are deprecated."
```
whilo: I find `warn` a bit stark, `info` or `debug` should be fine. We should only do this runtime warning if we really have to change behaviour. If vector and sequence arguments are supposed to be used for something else in the future then it would be reasonable to do so, but are they? I personally find it compact to just pass a sequence of maps or Datoms, in particular if you transact something tiny.

whilo: Ok, I have not made the decision to deprecate it. I guess it is best to ask @kordano and either send the message or keep it as is.

whilo: I talked to @kordano and we can keep the support for vector and sequence arguments, so I removed the deprecation message.

src/datahike/api.cljc
```diff
@@ -203,9 +203,7 @@
                                                   {:error :transact/syntax
                                                    :argument-keys (keys arg-map)}))
                 (or (vector? arg-map)
-                    (seq? arg-map))   (do (log/warn "Vector and sequence as argument are deprecated."
-                                                    "Instead use a map with key :tx-data.")
-                                          {:tx-data arg-map})
+                    (seq? arg-map))   {:tx-data arg-map}
```
TimoKramer, line 206: in api-ns there are docstrings where the deprecation is mentioned. that should be removed then as well.

TimoKramer: and the deprecation is also mentioned in connector.cljc line 145

whilo: I have improved the doc string for `api/transact`, but I don't really see where deprecation is indicated there. `connector.cljc:145` points to unrelated code for me. Which code did you mean?

jsmassa: I think, he is talking about the location where you copied this from.

TimoKramer: sorry, my bad.

jsmassa (commented):
Looks good to me now. I would suggest a review from @kordano though before merging.

whilo:
Ok, I just did a test also for the version checks we have now (to never load a database with older code versions than it was written with) and removed an error I accidentally applied as a wrong fix earlier.

src/datahike/writing.cljc
```diff
@@ -0,0 +1,239 @@
+(ns datahike.writing
```
kordano, line 1: Where do I find tests for this namespace?

kordano (commented):
Looks well written in total but it lacks testing. With such a large new use case we should have a decent amount of tests covering the new feature.

src/datahike/api.cljc
```diff
@@ -96,15 +102,20 @@
...
+              ;; By writing historical data in a separate index, datahike has the capability of querying data from any point in time.
               ;; You may control this feature using the `:keep-history?` attribute:
               (create-database {:store {:backend :mem :id \"example\"} :keep-history? false})
 
               ;; Initial data after creation may be added using the `:initial-tx` attribute, which in this example adds a schema:
               (create-database {:store {:backend :mem :id \"example\"} :initial-tx [{:db/ident :name :db/valueType :db.type/string :db.cardinality/one}]})"}
 
   create-database
-  dc/create-database)
+  (fn [& args]
```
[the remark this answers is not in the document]
whilo: Because this namespace overall is written that way. I can change that.

src/datahike/config.cljc
```diff
@@ -158,17 +165,17 @@
                  :initial-tx (:datahike-intial-tx env)
                  :keep-history? (bool-from-env :datahike-keep-history *default-keep-history?*)
                  :attribute-refs? (bool-from-env :datahike-attribute-refs *default-attribute-refs?*)
-                 :name (:datahike-name env (or *default-db-name* (z/rand-german-mammal)))
```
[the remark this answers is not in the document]
whilo: I wrote to you about in January or so. The problem is that I cannot de-duplicate the configuration if it is randomized. If `zufall` could be seeded with a random seed than it would be possible. I did not find dependencies on the `:name` feature so I decided to remove it. Note that I added `store-identity` instead, which fully identifies each database through the store it is based on. It is human readable and informative, but not as fun as `zufall` names.

src/datahike/writing.cljc
whilo: This namespace is just factored out from the prior `connection` namespace. It is tested with all the existing tests, I just tried to factorize the code base a bit better as `connection` was conflating too many things.

src/datahike/writing.cljc
```diff
@@ -0,0 +1,239 @@
...
+            [datahike.config :as dc]
+            [konserve.core :as k]
+            [taoensso.timbre :as log]
+            [hasch.core :refer [uuid]]))
+
+;; mapping to storage
+
+(defn stored-db? [obj]
+  ;; TODO use proper schema to match?
+  (let [keys-to-check [:eavt-key :aevt-key :avet-key :schema :rschema
```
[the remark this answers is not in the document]
whilo: > Looks well written in total but it lacks testing. With such a large new use case we should have a decent amount of tests covering the new feature.

I agree, but note that the code does not change the way the connection works for existing use cases and hence is tested for them with the current code base. Accessing a remote connection should be tested with an integration test including `datahike-server` and `datahike-server-transactor`. Since we also want to show case that, it would be good for me to get some feedback how to do it. We can then add an integration test in the same way we do it for datahike with konserve. Is that ok? I still think that this can be merged now then and the additional test can be added with the new use case.

kordano:
Yeah, it's fine to add them later on. So go ahead an release it. My other comments where primarily superficial and about formatting.

jsmassa (approved):

PR #333, opened 2021-05-21
Array support

### LINKED ISSUE #314: blobs / unstructured data
Is there a way to add a data blob / or fields that are not defined in the schema?
If there is no searching for this fields, this is no problem.

whilo:
Add (byte) array support with an ordering based on shape and value. Adding support for more array types is easy to add. This PR defines a total ordering over byte arrays that is first by shape (shorter is smaller) and then by element-wise Clojure comparison semantics. This allows pattern matching on binary data while avoiding deep comparisons on objects of different form. Resolves #314.

Different types of indexing can be beneficial depending on the domain. This is also an example of how a new sequential binary type can be added to Datahike.

### DIFF
--- a/src/datahike/array.cljc
+++ b/src/datahike/array.cljc
@@ -1,0 +1,32 @@
+(ns datahike.array
+  (:require [hitchhiker.tree.key-compare :as kc]
+            [hitchhiker.tree.node :as n]))
+
+(defn compare-arrays [a b]
+  ;; TODO can avoid some of these safety checks?
+  (def foo [a b])
+  (if (not (and (bytes? a) (bytes? b)))
+    (try
+      (compare a b)
+      (catch ClassCastException _
+        (- (n/-order-on-edn-types a)
+           (n/-order-on-edn-types b))))
+    (let [lc (compare (alength a) (alength b))]
+      (if (not (zero? lc))
+        lc
+        (loop [step (alength a)]
+          (if-not (zero? step)
+            (let [ec (compare (aget a (dec step)) (aget b (dec step)))]
+              (if (not (zero? ec))
+                ec
+                (recur (dec step))))
+            0))))))
+
+(defn a=
+  "Extension of Clojure's equality to things we also want to treat like values,
+  e.g. certain array types."
+  [a b]
+  (or (= a b)
+      (and (bytes? a)
+           (bytes? b)
+           (zero? (compare-arrays a b)))))
--- a/src/datahike/array.cljc
+++ b/src/datahike/array.cljc
@@ -1,10 +1,8 @@
 (ns datahike.array
-  (:require [hitchhiker.tree.key-compare :as kc]
-            [hitchhiker.tree.node :as n]))
+  (:require [hitchhiker.tree.node :as n]))
 
 (defn compare-arrays [a b]
   ;; TODO can avoid some of these safety checks?
-  (def foo [a b])
   (if (not (and (bytes? a) (bytes? b)))
     (try
       (compare a b)
--- a/src/datahike/array.cljc
+++ b/src/datahike/array.cljc
@@ -1,24 +1,33 @@
 (ns datahike.array
   (:require [hitchhiker.tree.node :as n]))
 
-(defn compare-arrays [a b]
-  ;; TODO can avoid some of these safety checks?
+(defn compare-arrays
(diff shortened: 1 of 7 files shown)


whilo:
@f-f @hellerve What do you think about defining an order on byte-arrays? To retrieve values from the database the most useful feature is value based equality, I think. This is why my first iteration of this PR orders first by length of arrays and only if they have equal lengths we do a more expensive element-wise comparison. This is a bit annoying if you expect prefix based orderings though, which for many data formats could induce a meaningful order (if your array dimensions are ordered in the right way). So I decided now to define an order that is similar to strings, but on byte level.

Datomic does not provide any value semantics for arrays, https://docs.datomic.com/on-prem/tech-notes/querying-byte-array.html. I think this is a bit unnecessary and a superficial buy-in into Java semantics, even though it can be expensive to do comparisons on arrays that share large prefixes (e.g. file format headers). Arrays should be treated similarly to strings, I think.

hellerve:
I think prefix ordering makes sense; it cuts out the possibility of early termination on length mismatch, that is true, but other than that (which is a pretty big condition, admittedly), it doesn’t need to be much slower at all.

Just a few ideas here to optimize:

- If we wanted to impose string semantics, but not care to re-implement the hopefully efficient string comparison ourselves, we could just cast the bytes to strings. This is inelegant to say the least, but ensures consistency and is a rather braindead solution (which is good, since it doesn’t incur a maintenance cost).
- The second best solution would be to pull in something like `memcmp`, which matches the semantics. I’m not aware of any possibilities of using this in Clojure or Java, though, which is a bummer.
- The third best solution is optimizing the hell out of the loop. For the sake of argument, here is my unbenchmarked braindump:

```clojure
(defn compare-arrays [a b]
 ; this should stay
 (if (not (and (bytes? a) (bytes? b)))
    (try
      (compare a b)
      (catch ClassCastException _
        (- (n/-order-on-edn-types a)
           (n/-order-on-edn-types b))))
      ; here we just precompute to save some comparisons
      (let [bl (alength b)
            al (alength a)
            end (min al bl)
            on-end
              (cond
                (> al bl) 1
                (< al bl) -1
                :else 0)
          ]
        (loop [i 1]
          (if (> i end)
            on-end
            ; to play around with efficiency, we could investigate whether
            ; (let [ae (aget a (dec i))
            ;       be (aget b (dec i))]
            ;   (if (= ae be)
            ;     (recur (inc i))
            ;     (- ae be)))
            ; is a little faster.
            (let [ec (compare (aget a (dec i)) (aget b (dec i)))]
              (if (zero? ec)
                (recur (inc i))
                ec)))))))
```

whilo:
@hellerve This is great feedback, thank you! Yes, I was mostly thinking of the things people might dump into arrays, e.g. small images (user images, icons etc.), in this case comparisons make no sense and the length mismatch would allow us to partition the space much more efficiently. But then for some array formats the prefix matching might be handy and allow efficient range scans (do you know of good examples?). Notably we could still change this later if we only guarantee equality matches for now as this can work the same for both orders and leave the prefix matching as an experimental feature. 

@hellerve Your code was not really faster in my microbenchmarks (the JVM typically optimizes these things well), but I looked into other options of using runtime primitives and it turns out that `java.util.Arrays/compare` was recently added and does the job very well and is orders of magnitude faster and better than casting to strings (which I tried before). (30 ms vs. 20 secs on a million bytes...)

@TimoKramer The CI/CD is failing because we use OpenJDK 8 and the Array comparators were introduced in OpenJDK 9. Can we bump the versions in our CI/CD? https://javabydeveloper.com/comparing-arrays-java-9/

TimoKramer:
@whilo setting the build JVM is no problem. The problem is more that people using Datahike with JDK8 will have problems. Last time we published Datahike built with JDK > 8 there were people complaining pretty soon. But JDK8 has EOL end of March 2022 so at some point we need to switch. If there is good reason to go to JDK 11 I am in favor of it but we need to talk it through with @kordano because he probably knows of companies using Datahike with JDK8.

f-f (dismissed):
Good discussion - using `java.util.Arrays.compare` is a great idea :clap:

whilo:
This should now be mergeable.

jsmassa (approved):
Looks good to me. Thanks for implementing this :)

PR #340, opened 2021-06-02
Comparators

jsmassa:
Closes issue #328

### DIFF
--- a/src/datahike/query.cljc
+++ b/src/datahike/query.cljc
@@ -202,47 +202,118 @@
   (reduce (fn [a b]
             (if b (reduced b) b)) nil args))
 
-(defmulti -lesser?
-  {:arglists '([value & more])}
-  (fn [value & more] (class value)))
+(defprotocol CollectionOrder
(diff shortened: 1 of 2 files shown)


jsmassa:
Tests are still failing on this due to arity exceptions. I cannot figure out why, but it seems to have to do with the protocols.

kordano:
Could you post the stacktrace?

jsmassa:
I fixed it now. Apparently it is not possible to give an arbitrary number of arguments to a protocol function.

whilo (approved):
This looks good. The cljs version is incomplete (you provide implementations for `java.util.Date`, which does not exist in cljs), but we can fix this later also. Also, for the data implementation, is it faster compared to the default reduction because you add the type hints?

whilo:
Wasn't the original goal to add a string comparator?

jsmassa:
Yes, and so I did. But it was that general that it turned out to work with all types so far, so I made it the default. I only left the special implementations that were already there.

PR #341, opened 2021-06-04
Add search cache

jsmassa:
This cache allows the user to restrict the amount of datoms being cached

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -2,6 +2,7 @@
   (:require
    #?(:cljs [goog.array :as garray])
    [clojure.walk]
+   [clojure.core.cache.wrapped :as cw]
    [clojure.data]
    #?(:clj [clojure.pprint :as pp])
    [datahike.array :refer [a=]]
@@ -10,8 +11,10 @@
    [datahike.constants :as c :refer [ue0 e0 tx0 utx0 emax txmax system-schema]]
    [datahike.tools :refer [get-time case-tree raise]]
    [datahike.schema :as ds]
+   [datahike.lru :refer [lru-datom-cache-factory]]
    [me.tonsky.persistent-sorted-set.arrays :as arrays]
    [datahike.config :as dc]
+   [environ.core :refer [env]]
    [clojure.spec.alpha :as s]
    [taoensso.timbre :refer [warn]])
   #?(:cljs (:require-macros [datahike.db :refer [defrecord-updatable cond+]]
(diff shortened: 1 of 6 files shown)


src/datahike/constants.cljc
```diff
@@ -5,3 +5,4 @@
 (def ^:const emax 0x7FFFFFFF)
 (def ^:const txmax 0x7FFFFFFF)
 (def ^:const implicit-schema {:db/ident {:db/unique :db.unique/identity}})
+(def ^:const max-db-caches 5)
```
whilo, line 8: Shouldn't this be a configurable parameter?

whilo (changes requested):
Very nice work. I think it would be good to have a way to monitor the caching behaviour, in particular when elements are added and removed. This can be done with logging, but I think it is good to make it easy to have diagnostics also in production systems (not just for this cache).

src/datahike/constants.cljc
jsmassa: Yes - maybe by environment variable? 
It can't be given with the database config since it is a global parameter.

src/datahike/db.cljc
```diff
@@ -677,6 +677,70 @@
 
 ;; ----------------------------------------------------------------------------
 
+(def db-caches (atom {}))
+(def db-cache-sizes (atom {}))
+
+(defn insert-cache [db-hash cache]
+  (let [caches @db-caches]
+    (when (>= (count caches) max-db-caches) ;; TODO: better part of record so it can be destroyed?
+      (let [last-used-db (first (reduce (fn [[d-hash1 {^Date d1 :last-access :as cache1}]
```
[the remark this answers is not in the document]
jsmassa: This is outdated now, but I think you are right

src/datahike/constants.cljc
whilo: Could we do it on a database id level (not value, but uuid that is globally fixed for each db)? I think @kordano has not added that yet, but in the longer run it might be good to be able to decide how big the cache should be for individual databases.

jsmassa: That would be a good idea. Would make it more intuitive than defining it only per database snapshot, which is how it works at the moment

whilo (dismissed):

whilo:
I decided that this is ok even as it is to go in, we can make it more configurable later as well, if this is too much work for you now @jsmassa. It would be nice to have control over how many databases can be cached though.

whilo (dismissed):
This looks good. Nice work!

whilo (approved):

Issue #345, opened 2021-06-21
allow `:db/isComponent` and `:db/noHistory` to be changed


kordano:

MrEbbinghaus:
`:db/doc` as well!

Every attribute not in the `db` namespace should be allowed.
`:my/custom-attribute`

PR #352, opened 2021-06-30
Allow more schema attributes to be updated

MrEbbinghaus:
This MR allows the update of any non-`db` namespaced attribute key

- [x] ~~any non-`db` namespaced attribute key~~ (There are other places that need changes, other than the validation fn)
- [x] The `:db/doc` key
- [x] The `:db/isComponent`key
- [x] The `:db/noHistory` key

Closes https://github.com/replikativ/datahike/issues/345

### DIFF
--- a/src/datahike/schema.cljc
+++ b/src/datahike/schema.cljc
@@ -167,14 +167,18 @@
        (when (and (not= old-value new-value) 
                   (= "db" (namespace attr-def)))
          (case attr-def
-           :db/cardinality (if (= new-value :db.cardinality/many)
-                             (if (get-in attr-schema [:db/unique])
-                               (assoc m attr-def [old-value new-value])
-                               nil)
-                             (assoc m attr-def [old-value new-value]))
-           :db/unique (when-not (get-in attr-schema [:db/unique])
-                        (when-not (= (get-in attr-schema [:db/cardinality]) :db.cardinality/one)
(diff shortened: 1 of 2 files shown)


MrEbbinghaus:
@kordano I don't think there is anything preventing `:db/cardinality` or `:db/noHistory` from being updated, at least as far as I would assume with my sparse knowledge of the Datahike implementation. What do you think?

kordano:
Both changes are fine. There might be some old history data when you set `:db/noHistory` but this should be transparent the database engineer when looking at the schema changes.

kordano:
And thanks for the PR, happy to review as soon as you like. :)

MrEbbinghaus:
@kordano Added the two attributes as well. Feel free to review.

kordano (changes requested):
Looks good @MrEbbinghaus. Could you add a small test that asserts changing `:db/doc`, `:db/noHistory`, and `:db/isComponent` to the `datahike.test.schema` test namespace?

MrEbbinghaus:
Ok, will do it on the weekend

MrEbbinghaus:
@kordano I removed the check to allow non `db` namespaced attributes, as that would involve other parts of the code to be changed. Maybe this can be tackled in a follow-up PR.

kordano (approved):
This looks good. Thanks for your contribution @MrEbbinghaus.

PR #357, opened 2021-07-07
many minor adjustments for ClojureScript

timothypratley:
This patch doesn't solve everything.
It does get significantly further in the compilation (builds with warnings),
but more work would be required to make it actually work correctly.
Suggestions/collaboration sought on achieving full compatibility.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,5 +1,5 @@
 {:deps {org.clojure/clojure                         {:mvn/version "1.10.1"}
-        org.clojure/clojurescript                   {:mvn/version "1.10.597"}
+        org.clojure/clojurescript                   {:mvn/version "1.10.773"}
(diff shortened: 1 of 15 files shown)


kordano:
Thanks for your contributions @timothypratley. Currently we're adjusting konserve for sync and async versions as preparations for the cljs port. Maybe @whilo and @groundedSAGE could have a look.

timothypratley:
You're welcome! Thanks for the heads up about Konserve.

whilo:
@timothypratley Thanks for your work! The rebase of the cljs support is planned after we have landed the current changes for scaling out in our 1.0 release. I assume you are aware of the `cljs-support` branch. We will not do much more work on it, but reapply the changes there in a slightly different way to the current development branch, so your work here is definitely valuable. We are very happy to discuss collaborating on it if you are interested (best on Discord).

Issue #353, opened 2021-06-30
improve paginate function


kordano:
- refactor current implementation
- improve performance

PR #360, opened 2021-07-10
Improve pagination performance

jsmassa:


jsmassa:
This version doesn't show drawbacks for other cases while still improving pagination immensely.

Closes Issue #353

kordano (approved):

PR #362, opened 2021-07-12
Improve upsert

grischoun:
- Fix a bug illustrated by the added tests.
- Should slighty improve upsert performance, especially on temporal indices.


whilo (approved):
This looks good to me.

jsmassa (approved):
Nice work :)

PR #365, opened 2021-07-19
363 fix history duplicates

kordano:


src/datahike/db.cljc
```diff
@@ -367,34 +368,62 @@
...
+  ([[f & others :as fns] order a b]
+   (if (seq fns)
+     (let [result (kc/-compare (f a) (f b))
+           f-result (if (= order >) (* -1 result) result)]
+       (if (= 0 f-result)
+         (recur others order a b)
+         f-result))
+     0)))
+
+(defn concat-sort [list-a list-b index-type]
```
grischoun, line 384: How efficient is this fn?

grischoun (dismissed):
Except the minor comments, it looks very good.

src/datahike/datom.cljc
```diff
@@ -103,7 +103,9 @@
 (defn- equiv-datom [^Datom d ^Datom o]
   (and (== (.-e d) (.-e o))
        (= (.-a d) (.-a o))
-       (= (.-v d) (.-v o))))
+       (= (.-v d) (.-v o))
+       (= (.-tx d) (.-tx o))
```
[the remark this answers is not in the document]
whilo: I also find the addition of tx surprising, I think then it is just equality and not equivalence, right?

src/datahike/db.cljc
whilo: The algorithmic complexity of this algorithm is the length of the longest list in {list-a, list-b}, which is better than resorting everything (which is O(n*log n) and optimal here. If all of the output is expected to be consumed it can be more efficient to remove the laziness, but I guess sometimes only part of it will be consumed, right?

src/datahike/db.cljc
```diff
@@ -367,34 +368,70 @@
...
+           (cons (first list-a) (lazy-seq (merge-sorted comp (rest list-a) (rest list-b))))
+           (cons (first list-b) (lazy-seq (merge-sorted comp list-a (rest list-b))))))))))
+
+(defn concat-sort [list-a list-b index-type]
+  (let [order (case index-type
+                :eavt [:e :a :v :tx :added]
+                :aevt [:a :e :v :tx :added]
+                :avet [:a :v :e :tx :added]
+                [:e :a :v :tx :added])]
+    (vec (merge-sorted #(multi-comp order < %1 %2) list-a list-b))))
```
whilo, line 402: Here it is not lazily consumed, because `vec` realizes the full vector.

whilo (commented):
Nice work!

jsmassa:
The temporal searches now first return the elements only in the current index, then everything else from the temporal index. This avoids having duplicates but does not add more complexity by having to perform a distinct operation. 

Note that - as before - the datoms are not being sorted.

src/datahike/db.cljc
```diff
@@ -1560,20 +1557,23 @@
...
-      (let [e      (:e datom)
-            v      (if (datom-added datom) (:v datom) nil)
-            queue  (or (-> report' ::queued-tuples (get e)) {})
-            tuples (get (-attrs-by db :db/attrTuples) a)
-            queue' (queue-tuples queue tuples db e v)]
-        (update report' ::queued-tuples assoc e queue'))
-      report')))
+(defn- transact-report
+  ([report datom] (transact-report report datom false))
+  ([report datom upsert?]
```
kordano, line 1562: Why do you use this polymorphism approach instead of multimethods or protocols?

src/datahike/datom.cljc
```diff
@@ -101,7 +101,7 @@
       (combine-hashes (hash (.-v d)))))
 
 (defn- equiv-datom [^Datom d ^Datom o]
-  (and (== (.-e d) (.-e o))
+  (and (== (.-e d) (.-e o)) ;; why ==?
```
kordano, line 104: We inherited this from DataScript.

src/datahike/db.cljc
jsmassa: In this case I was just looking for a simple way to avoid code duplication to make it clearer what is happening for upserts opposed to inserts. Protocols would not have been applicable since we are not dispatching on type and multimethods would have lead to code duplication again, which I wanted to avoid in the first place. 

For the `with-datom` function we could think about dispatching on the operations `:insert` or `:upsert`. This would make things clearer I'd say.

src/datahike/datom.cljc
jsmassa: Apparently it's the type independent version of `=`. Seems counterintuitive to me though that `==` is a weaker equality than `=`.
I'd say `==` makes sense here though. I will remove the comment.

whilo: I am not sure why `==` is needed here. This only makes sense if the entity-id provided could sometimes be a floating point type on the JVM. In JS there is only one number type.

jsmassa: Do you think, we should should just use `=` then?
When I think about it, it would make sense to me since in other places - e.g. the `cmp-datoms-eavt` function etc. - the compiler fails if one of the entity ids is not of type `Integer`.

jsmassa: Addressed in issue #385

jsmassa (approved):
All questions have been addressed now

PR #366, opened 2021-07-23
364 cardinality many duplicate datoms

grischoun:
Fix the duplicate issue when inserting on cardinality many attributes.


grischoun:
I still need to check why two tests fail.

grischoun:
This is now ready for review.

There is potential for improving 'retraction' performance, I think. I'll investigate more on that later.

src/datahike/index/hitchhiker_tree/insert.cljc
```diff
@@ -0,0 +1,62 @@
...
+    (if (exists-old? kvs key)
+      kvs
+      (assoc kvs key nil)))
+  (-apply-op-to-tree [_ tree]
+    (let [children (cond
+                     (tree/data-node? tree) (:children tree)
+                     :else (:children (peek (tree/lookup-path tree key))))]
+      (if (exists-old? children key)
+        tree
+        (tree/insert tree key nil)))))
```
jsmassa, line 47: How does this work so there are no duplicates when the key contains the transaction number? 
Don't you test for datoms from 2 different transactions? So how could the old key be found? 
I assume I am overlooking something here...

grischoun: Good catch @jsmassa !
It now replaces the old key with the new one (like with the persistent-set index).

jsmassa (approved):
Nice, then I am happy now with the code :)

PR #369, opened 2021-08-09
Add transaction log

jsmassa:


test/datahike/test/api.cljc
```diff
@@ -672,3 +672,115 @@
                   hash-2 (hash @conn)]
               (is (not= hash-1 hash-2))
               (is (not= hash-0 hash-2)))))))))
+
+(defn test-write-schema [id cfg]
```
grischoun, line 676: These tests are not related to directly testing the introduction of `tx-log`, right? They are here to test that what existed before is not broken, right?

grischoun (approved):
Sounds great!
Good job!

kordano:
will be implemented with #377

PR #373, opened 2021-08-20
Fix translations

jsmassa:


test/datahike/test/attribute_refs/db.cljc
```diff
@@ -0,0 +1,41 @@
...
+   :keep-history? true
+   :attribute-refs? true
+   :schema-flexibility :write
+   :name "attr-refs-test"})
+
+(def name-schema [{:db/ident :name
+                   :db/cardinality :db.cardinality/one
+                   :db/valueType :db.type/string}])
+
+(defn setup-db [cfg]
```
kordano, line 18: At some point we could move this setup into a `test/utils` namespace.

kordano (approved):

PR #383, opened 2021-08-28
Test read handlers for inserts

grischoun:


grischoun:
@whilo Don't we have to add an 'add-insert-handler' here https://github.com/replikativ/datahike/blob/ca18bd63849c2ce3f59f34591ce651ad22efc922/src/datahike/store.cljc#L100?

whilo (approved):
Good job!

PR #390, opened 2021-09-12
342 refactor test namespaces

TimoKramer:
The tuples-tests are failing because of changes in behaviour of datoms-fn in api-ns. Can you verify the intended behaviour and give me a hint on how to rewrite the tests?


test/datahike/test/transact_test.cljc
```diff
@@ -0,0 +1,394 @@
...
+  @(d/transact! conn {:tx-data [[:db/add 1 :name "Petr"]]})
+  @(d/transact! conn {:tx-data [[:db/add 1 :aka  "Devil"]]})
+  @(d/transact! conn {:tx-data [[:db/add 1 :aka  "Tupen"]]})
+  (d/q '[:find ?v
+         :where [?e :aka ?v]] @conn)
+
+  (d/datoms @conn {:index :eavt}))
+
+(deftest test-db-fn
+  (let [conn (du/setup-db {:initial-tx [{:db/init :aka :db/cardinality :db.cardinality/many}]})
```
kordano, line 254: The intial tx should be `{:db/ident :aka :db/cardinality :db.cardinality/many}`

TimoKramer:
Please wait with a merge, review would be good though

TimoKramer:
Should be good to go now :+1:

test/datahike/test/explode_test.cljc
```diff
@@ -1,41 +1,49 @@
...
-                      :where [1 :name ?n]
-                      [1 :age ?a]] @conn)
-               #{["Ivan" 16]}))
-        (is (= (d/q '[:find  ?v
-                      :where [1 :also ?v]] @conn)
-               #{["ok"]}))
-        (is (= (d/q '[:find  ?v
-                      :where [1 :aka ?v]] @conn)
-               #{["Devil"] ["Tupen"]}))))))
+#_(deftest test-explode
```
kordano, line 13: Why is this excluded?

test/datahike/test/utils.cljc
```diff
@@ -1,7 +1,50 @@
 (ns datahike.test.utils
-  (:require [datahike.api :as d]))
+  (:require [datahike.api :as d])
+  (:import (java.util UUID)))
 
-(defn setup-db [cfg]
-  (d/delete-database cfg)
-  (d/create-database cfg)
-  (d/connect cfg))
+(defn deep-merge
```
kordano, line 5: We already have another `deep-merge` in the [config ns](https://github.com/replikativ/datahike/blob/7e60af807dd4db1d0eb73b75ac2f010e31361a3a/src/datahike/config.cljc#L85), but this here looks more sophisticated. Maybe we could move this to `datahike.tools` and use it in `datahike.config`?

kordano (changes requested):

tests.edn
```diff
@@ -4,4 +4,5 @@
                        :type :kaocha.type/cljs
                        :ns-patterns ["datahike.test."]}
                     {:id :integration
-                     :test-paths ["test/datahike/integration_test"]}]}
+                     :test-paths ["test/datahike/integration_test"]}]
+            :reporter kaocha.report/documentation}
```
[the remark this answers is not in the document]
TimoKramer: I can remove it, it is just a command line option. you could probably overwrite it again with something like `--reporter kaocha.report/dots` everytime. I have no strong opinion on that.

test/datahike/test/utils.cljc
TimoKramer: good point :t-rex:

kordano (approved):

tests.edn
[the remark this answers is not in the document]
kordano: Yeah, I'll just use that locally then when I'm running the tests. I don't need the names of all the tests but only the ones that are failing.

Issue #388, opened 2021-09-08
Equivalent datoms should only be added once to database


jsmassa:
Due to compatibility with Datomic. 
Datoms with the same e-a-v triple should be omitted and not added to the database again with a different transaction ID.

PR #392, opened 2021-09-14
388 no reinsertion of equiv datom

grischoun:
Insertion of the same datom shall not replace the original one.
This is to preserve the same semantics as Datomic.
Note that the 'mem-set' backend does not behave this way though.

### DIFF
--- a/src/datahike/index/hitchhiker_tree/insert.cljc
+++ b/src/datahike/index/hitchhiker_tree/insert.cljc
@@ -18,7 +18,6 @@
           true
           indices))
 
-
 (defn exists-old?
   "Returns true if 'new' already exists in 'old-keys'."
   [old-keys new]
--- a/src/datahike/index/hitchhiker_tree/insert.cljc
+++ b/src/datahike/index/hitchhiker_tree/insert.cljc
@@ -18,39 +18,34 @@
           true
           indices))
 
-(defn old-key
-  "Returns the key with the same values as 'new' at the positions given by 'indices', if it exists."
-  [kvs new indices]
-  (when (seq kvs)
-    (let [mask (mask new indices)]
-      (when-let [candidate (->> (subseq kvs >= mask)
-                                (map first)
-                                first)]
-        (when (equals-at-indices? indices candidate new)
-          candidate)))))
 
-(defn remove-old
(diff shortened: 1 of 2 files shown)


whilo:
The code changes themselves make sense to me, but I lack a bit of context. #388 does not provide a clear description of the problem to me. How does this affect queries for instance?

Also, what would be necessary to keep the persistent set compatible? I think subtle mismatches in the semantics should be avoided, but I understand that it might be tricky.

grischoun:
> The code changes themselves make sense to me, but I lack a bit of context. #388 does not provide a clear description of the problem to me. How does this affect queries for instance?
> 
> Also, what would be necessary to keep the persistent set compatible? I think subtle mismatches in the semantics should be avoided, but I understand that it might be tricky.

This PR actually reverts the semantics to what it was in previous versions of Datahike. So query performance should not change and I find it safer to not change the persistent set semantics because that's what users are used to so far. 

(A change in semantics is exactly the reason why this PR exists.)

src/datahike/index/hitchhiker_tree/insert.cljc
```diff
@@ -18,39 +18,33 @@
...
+(defn exists-old?
+  "Returns true if 'new' already exists in 'old-keys'."
+  [old-keys new]
+  (when (seq old-keys)
+    (let [indices [0 1 2]
+          mask (mask new indices)]
+      (when-let [candidates (subseq old-keys >= mask)]
+        (->> candidates
+             (map first)
+             first
```
jsmassa, line 30: Whats the reason for not using `ffirst`?

jsmassa (dismissed):
Does what it should for the hitchhiker tree.

I agree with @whilo though that the persistent set index should react the same as the hitchhiker-tree index. I know that trying to do that got us in trouble in the first place, but I don't think that datascript compliance is as important as datomic compliance. We have to make a conscious decision about that though, maybe even document the differences if we derive from the original datascript behavior. 

For now I think changing the behavior of the hitchhiker-tree index to the previous behavior is enough because it fixes the current issue. We can still change it for the persistent set in a separate PR.

kordano (approved):

PR #414, opened 2021-11-10
Release 0.4.0

kordano:
#### SUMMARY
Release 0.4.0


bin/run-integrationtests
```diff
@@ -19,7 +19,7 @@ function teardown() {
 
 CONTAINER_NAME=$(setup)
 
-bin/kaocha --focus :integration
+TIMBRE_LEVEL=':fatal' bin/kaocha --focus :integration
 
-DATAHIKE_STORE_BACKEND=pg DATAHIKE_STORE_HOST=localhost DATAHIKE_STORE_PORT=5432 DATAHIKE_STORE_USER=alice DATAHIKE_STORE_PASSWORD=foo DATAHIKE_STORE_DBNAME=config-test bin/kaocha --focus datahike.integration-test.config-record-test
+# DATAHIKE_STORE_BACKEND=jdbc DATAHIKE_STORE_DBTYPE=postgresql DATAHIKE_STORE_HOST=localhost DATAHIKE_STORE_PORT=5432 DATAHIKE_STORE_USER=alice DATAHIKE_STORE_PASSWORD=foo DATAHIKE_STORE_DBNAME=config-test bin/kaocha --focus datahike.integration-test.config-record-test
```
jsmassa, line 24: Is it a comment on purpose for now? 
I am also not sure how environment variables work together with external stores at moment since I saw two different ways of doing it and I couldn't find the spot in the code where the variables are read.

jsmassa (changes requested):
Except for the two minor things that I addressed, it looks good to me.
 
I would maybe also delete the query-v3 stuff though since it is code that we are not using and if we wanted to we could still retrieve it from the DataScript repo. It might be beneficial to talk to Nikita about what he was trying to do there at some point, i.e. where he saw weaknesses about the current query code, but this is not pressing for this release.

bin/run-integrationtests
kordano: Removed it, since I'm using a fixed postgres integration config. I'll create a bug report for the missing env config for the nested jdbc config.

jsmassa (approved):
Thanks:)

Issue #440, opened 2021-11-23
[Bug]: store max-ent-id in db record


whilo:
### What version of Datahike are you using?

0.3.7-SNAPSHOT

### What version of Java are you using?

openjdk 11.0.11 2021-04-20

### What operating system are you using?

Ubuntu 20.04

### What database EDN configuration are you using?

-

### Describe the bug

When max-ent-id is not stored in the db record the unit tests in the GraalVM branch fail.

### What is the expected behaviour?

Max-ent-id should not be decremented in these tests.

### How can the behaviour be reproduced?

Look at https://github.com/replikativ/datahike/pull/337

kordano:
Fixed with #458

PR #458, opened 2021-12-17
store max-eid as well

kordano:
#### SUMMARY
See #440 

##### Feature
- [x] Formatting checked
- [x] `CHANGELOG.md` updated


whilo (approved):
Looks good. Thank you!

whilo:
Oh, I forgot to think about this in the review, but I think you also need to still have a fallback to recover the `max-eid` when it is not yet stored in the database, i.e. extract it from the index again with `(db/init-max-eid eavt-key)` in this case, right?

Issue #467, opened 2022-01-25
[Bug]: konserve versions between release and testing are different


kordano:
### What version of Datahike are you using?

0.4.1480

### What version of Java are you using?

openjdk version "11.0.13" 2021-10-19

### What operating system are you using?

Ubuntu 20.04.3 LTS

### What database EDN configuration are you using?

Any

### Describe the bug

The used konserve version differs between different builds.

### What is the expected behaviour?

Testing should have the some konserve, hitchhiker-tree and clojure dependencies as the release build.

### How can the behaviour be reproduced?

Run a repl with `dev`and `test` build, check Datahike versions with `:meta` in the database record. Then see the difference:
`dev` is using konserve `0.5.3`while `test` is using `0.6.0alpha-3`.

yflim:
I reproduced the issue and got konserve `0.5.1` with `dev` and `0.6.0alpha-1` with `test`.

Dependency trees from `clj [-A:<test|dev>] -Stree` (and, for good measure, dependency traces from `clj [-A:<test|dev>] -Strace`) point to the following paths for konserve:

In all builds: `io.replikativ/hitchhiker-tree 0.1.11` -> `io.replikativ/konserve 0.5.1`

However, that is superseded in `test` by:
`io.replikativ/datahike-jdbc 0.1.2-SNAPSHOT` -> `alekcz/konserve-jdbc 0.1.0-SNAPSHOT` -> `io.replikativ/konserve 0.6.0-alpha1`

(The version numbers `0.5.1` and `0.6.0alpha-1` are corroborated by [hitchhiker-tree/deps.edn](https://github.com/replikativ/hitchhiker-tree/blob/master/deps.edn) and [konserve-jdbc/project.clj](https://github.com/alekcz/konserve-jdbc/blob/master/project.clj), so it looks like there's a typo in the instructions.)

Might be worth writing something to check for other such inconsistencies?

kordano:
Yes, it would be helpful to search for more inconsistencies.
AdditionalIy we could use a specific version for konserve instead of indirect dependency through the hitchhiker-tree. Also it would be a good idea to move the integration tests in the relevant backend-project. So we would only test for the file-  and in-memory backend inside Datahike. 
Could you prepare a PR for that @yflim?

yflim:
For the record, the following are other current dependency inconsistencies between builds:
{com.cognitect/transit-java
 {:test #:mvn{:version "1.0.343"},
[...]

PR #473, opened 2022-01-31
467 konserve versions between release and testing are different

yflim:
#### SUMMARY
Fix inconsistent konserve versions across builds, plus related tasks #467 

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked
- [ ] `CHANGELOG.md` updated

#### ADDITIONAL INFORMATION
1. Fix inconsistent konserve versions across builds, as stated in summary.
2. Check for and fix other such inconsistencies.
3. Integrate checks for dependency inconsistencies across builds.
4. Doesn't quite seem to fit with the other changes, but as mentioned in #467, retain only integration tests for file- and in-memory backends in Datahike, and move others.


yflim:
About checking for dependency inconsistencies: I think it could be done entirely in a Clojure `main` that invokes dependency tracing then analyses the trace files, that can be invoked as needed as a `deps.edn` alias. (Alternatively, the alias could invoke a shell script that traces dependencies and then calls Clojure, but that seems less cohesive.) It seems to make less sense as a CI job now that I've looked into that.

Does this sound reasonable? If so, where should the code go? Does something like `src/dependencies_check.clj` make sense?

kordano:
Like with `build.clj` in the base folder, I would just create the file there. The `src` folder should only contain things that are built into Datahike.

deps.edn
```diff
@@ -1,6 +1,7 @@
 {:deps {org.clojure/clojure                         {:mvn/version "1.10.3"}
         org.clojure/clojurescript                   {:mvn/version "1.11.4"}
         io.replikativ/hitchhiker-tree               {:mvn/version "0.1.11"}
+        io.replikativ/konserve                      {:mvn/version "0.6.0-alpha1"}
```
jsmassa, line 4: I can see how that can fix the problem with different versions. Did you check if that is the case?

I only wonder if it would be better to place it above the hitchhiker-tree dependency because that should also pull in konserve

yflim: Yes, I checked the versions selected for different builds.

As for dependency positioning, yes, hitchhiker-tree does pull in konserve. I compared the existing code and the change you suggested, but selected package versions were the same in both cases, so I'm keeping the status quo.

kordano: Could you then go with the latest release of konserve `0.6.0-alpha3`. Or would that break something?

kordano (changes requested):
This looks good. Please resolve the merge conflict and then we can release it.

kordano (approved):

PR #481, opened 2022-02-15
Get max-eid from index if not present in db

### LINKED ISSUE #465: [Bug]: init max-ent-id from index if needed

0.3.7-SNAPSHOT

openjdk 11.0.11 2021-04-20

Ubuntu 20.04

- 

### Describe the bug

The max-ent-id is not recovered automatically from the index when an old Datahike instance is being used. 

We should have a fallback to recover the max-eid when it is not yet stored in the database, i.e. extract it from the index again with (db/init-max-eid eavt-key) in this case.

-

MrEbbinghaus:
#### SUMMARY
This fixes #465 by getting the max-eid from the index if it is not present in the db record.

##### Bugfix
- [ ] Integration tests added
   - I am not sure how to add this. 
- [ ] ~~Architecture Decision Record added if design changes necessary~~
- [x] Formatting checked
- [x] `CHANGELOG.md` updated
   - The same as: "Persist max-eid"?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Switch to GitHub Flow and using main branch
 - Switch to tools.build for building and deploying
 - Persist max-eid
+- Init max-eid from index for older db versions https://github.com/replikativ/datahike/issues/465 https://github.com/replikativ/datahike/pull/481
 - Allow attribute access to historical db records
 - Allow keyword keys for queries
 
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -234,6 +234,16 @@
                   (filter (fn [^Datom d] (= tx (datom-tx d))) (-all eavt)) ;; _ _ _ tx
                   (-all eavt)]))))
 
+(defn init-max-eid [eavt]
+  ;; solved with reserse slice first in datascript
+  (if-let [datoms (-slice
+                    eavt
+                    (datom e0 nil nil tx0)
+                    (datom (dec tx0) nil nil txmax)
+                    :eavt)]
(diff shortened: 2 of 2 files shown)


kordano:
Thanks for the contribution @MrEbbinghaus, I could draft a test snippet to test the max-eid. As for the changelog I would write something like initialize max-eid.

kordano (commented):
Thanks @MrEbbinghaus. For testing I'm creating a previous database that does not have the max-eid yet. I'll add the link to the database here once I have created it and made it available for testing.

MrEbbinghaus:
Hey @kordano tested it locally with a database of mine (Unfortunately I can't publish it).

I was thinking about an automated test for this?

MrEbbinghaus:
@kordano Hm.. Maybe the `:max-eid` should be ensured on connect? What do you think?

jsmassa:
I also think it should be ensured on `-connect`. As far as I can see, in all other cases, the `max-eid` should already be present.

@kordano Did the underlying problem also show up in your compatibility tests? If yes, can they serve as a test for this issue? Otherwise it seems to me like it could get awkward testing this because it would require importing an old database.

MrEbbinghaus:
@jsmassa Maybe this should be part of some kind of migration system? 

```clojure
(let [conf ...]
  (d.migration/migrate! conf)
  (d/connect conf))
```

I don't know if this is feasible, as it would require that any datahike version could connect to and migrate a database created with any previous version of datahike.

jsmassa:
I'd say it could get a bit too complex for something integrated into datahike itself. What do you think, @kordano?

MrEbbinghaus:
I will open a new MR for this issue, where I add the `max-eid` on connect.

Issue #477, opened 2022-02-07
[Bug]: history migration not working


kordano:
### What version of Datahike are you using?

0.4.1480

### What version of Java are you using?

openjdk version "11.0.13" 2021-10-19

### What operating system are you using?

Ubuntu 20.04.3 LTS

### What database EDN configuration are you using?

```edn
{:store              {:backend :mem
                      :id      "export"}
 :keep-history?      true
 :schema-flexibility :write
 :attribute-refs?    false}
```

### Describe the bug

When exporting a database with historical data and importing it into a new database, the historical data like removal and original data are lost.

### What is the expected behaviour?

History should be imported as well.

### How can the behaviour be reproduced?

``` clojure
(require '[datahike.api :as d])
(require '[datahike.migrate :as m])

(def schema [{:db/ident       :name
              :db/cardinality :db.cardinality/one
              :db/index       true
              :db/unique      :db.unique/identity
              :db/valueType   :db.type/string}
             {:db/ident       :parents
              :db/cardinality :db.cardinality/many
              :db/valueType   :db.type/ref}
             {:db/ident       :age
              :db/cardinality :db.cardinality/one
              :db/valueType   :db.type/long}])

(def export-cfg {:store              {:backend :mem
                                      :id      "export"}
                 :keep-history?      true
                 :schema-flexibility :write
                 :attribute-refs?    false})

(do
  (d/delete-database export-cfg)
  (d/create-database export-cfg))

(def export-conn (d/connect export-cfg))

(d/transact export-conn schema)

(d/transact export-conn {:tx-data [{:name "Alice" :age  25}
                                   {:name "Bob" :age  35}]})

(d/transact export-conn {:tx-data [{:name    "Charlie"
                                    :age     5
                                    :parents [[:name "Alice"] [:name "Bob"]]}]})

(d/transact export-conn {:tx-data [{:name "Daisy" :age 20}]})
(d/transact export-conn {:tx-data [{:name "Erhard" :age 20}]})

(d/transact export-conn {:tx-data [[:db/retractEntity [:name "Erhard"]]]})

(datahike.migrate/export-db @export-conn "/tmp/dh.dump")

(def import-cfg {:store              {:backend :mem
                               :id      "import"}
          :keep-history?      true
          :schema-flexibility :write
          :attribute-refs?    false})

(do
  (d/delete-database import-cfg)
  (d/create-database import-cfg))

(def import-conn (d/connect import-cfg))

[...]

PR #496, opened 2022-03-08
477 history migration not working

yflim:
#### SUMMARY
Include historical datoms in `datahike.migrate` export, so history is not lost in migration as would happen currently. #477

##### Bugfix
- [ ] Integration tests added
- [x] Formatting checked
- [ ] `CHANGELOG.md` updated


yflim:
I'm benchmarking 3 approaches: based on [`wanderung.datahike`](https://github.com/lambdaforge/wanderung/blob/main/src/wanderung/datahike.clj), sorting using the [Scicloj Tablecloth API](https://github.com/scicloj/tablecloth/blob/master/src/tablecloth/api.clj), and vanilla Clojure sorting.

Benchmarking parameters used were backend and index type, consisting of (in order) `mem` i.e. persistent-set or hitchhiker-tree indices, or `file`; and number of entities, in this case 0, 1000, 10000, 100000, or 1000000. All the approaches triggered an out-of-memory error at the 1M-entity stage, though the Wanderung approach failed sooner, failing at the `persistent-set` stage, whereas the others failed at `hht`.

benchmark/src/benchmark/measure.clj
```diff
@@ -103,23 +126,95 @@
...
+  (def test-clj-figures
+    (get-measurements test-opts configs {:spec-fn-name "export-db-clj"
+                                         :make-fn-invocation ret-clj-export-db-fn}))
+  (def test-clj-figs-filtered (filter-export-figures test-clj-figures))
+  test-clj-figs-filtered
+
+  (alter-var-root #'datahike.api/connect
+                  (fn [f] (fn [& args] (println "instrumented") (apply f args))))
+
+  )
```
jsmassa, line 220: Should be removed from this place before merging. If you want to keep this code to make further development easier, you could move it to a sandbox namespace in the dev directory and if there's something that might be useful in the future, it should be integrated properly into the codebase (if not already done).

benchmark/src/benchmark/measure.clj
```diff
@@ -11,84 +12,106 @@
...
+(defn measure-transaction-times
+  [{:keys [iterations tx-entity-counts]} conn config entity-count datom-count]
+  (log/info (str "Measuring transaction times on database with config named '" (:name config) ", "
+                 (count c/schema) " attributes in entity, and " entity-count " entities in database"))
+  (for [_ (range iterations)    ;; sampling for given entity-count
+        tx-entities tx-entity-counts
+        _ (range iterations)]   ;; sampling for each tx-entities
+    (let [tx (vec (repeatedly tx-entities (partial c/rand-entity Integer/MAX_VALUE)))
+          timed-transact (timed (d/transact conn tx))]
+      (d/transact conn (mapv (fn [dat] [:db.fn/retractEntity (.-e dat)])
```
jsmassa, line 78: `retract` only cleans up the database if the history is disabled. With history, it should be `purge`. So far, we are not testing configurations with history yet, but we should do that at some point. To avoid confusion and incorrect values in the future, we should already consider both options here.

jsmassa (commented):
Thanks for cleaning up the benchmark namespaces a bit. In the future, please make things like that a separate PR though with an appropriate subject.

benchmark/src/benchmark/measure.clj
yflim: Yes, I put it there temporarily in case it turned out to be useful. It's just been removed.

benchmark/src/benchmark/measure.clj
```diff
@@ -11,84 +12,106 @@
...
+           (time-context-map (:t (timed (d/q query @conn)))
+                             config function entity-count datom-count details)))))))
+
+(defn measure-transaction-times
+  [{:keys [iterations tx-entity-counts]} conn config entity-count datom-count]
+  (log/info (str "Measuring transaction times on database with config named '" (:name config) ", "
+                 (count c/schema) " attributes in entity, and " entity-count " entities in database"))
+  (for [_ (range iterations)    ;; sampling for given entity-count
+        tx-entities tx-entity-counts
+        _ (range iterations)]   ;; sampling for each tx-entities
```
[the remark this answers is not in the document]
yflim: The first corresponds to sampling each config, while the second corresponds to number of entities in each transaction (when benchmarking transactions). In principle, the two numbers should be independent, but they coincided in the benchmarking code and I stuck with the status quo.

benchmark/src/benchmark/measure.clj
yflim: Oh, I put it there just in case it turned out to be useful. It's just been removed.

benchmark/src/benchmark/measure.clj
yflim: Ah, I see. I used `retract` to avoid deleting and re-creating the database when avoidable—does it matter whether `retract` or `purge` is used in this case?

yflim:
`export-db-clj` consistently performed far better than the other approaches, `export-db-wanderung` and `export-db-tc`. Moreover, (I've just noticed) the latter doesn't support byte arrays, which is likely a dealbreaker. So I've replaced `export-db` with `export-db-clj` in commit [33a8c1a](https://github.com/replikativ/datahike/pull/496/commits/33a8c1a03ca74ba1a08683d018521dee28cb8639).

benchmark/src/benchmark/measure.clj
jsmassa: At the moment, purge [only works on databases with history](https://github.com/replikativ/datahike/blob/97eb463b8c71c60f9ff261db795ee679e4e78767/src/datahike/db.cljc#L2078). IMO it would be more intuitive if it would work with both configurations, but I guess that is a discussion on its own.

benchmark/src/benchmark/measure.clj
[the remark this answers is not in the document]
jsmassa: I see where you are coming from, but in the original version the database changed between the iterations, so it made at least a little sense to have both repetitions there because there was actual sampling going on (although even then one admittedly could have been skipped). In fact, this shows that the first iteration-loop should go back to its original spot, because otherwise the benchmarking is done on a single database. The reason why it makes sense to test different databases at the moment are mainly the caches on different levels that will skew the benchmark results especially for queries. As I understand we will be working on making the ones we have an influence on configurable, so we will be able to bypass this problem. For now we should stick to the momentary solution even though it is slower.  What we can do though is to actually make those two iteration counts different parameters, so at least one parameter to configure the amount of times a database with the same number of entities is built and tested, and one parameter to configure the amount of times a specific function is tested on a database. This way we would still have the option to decide what is best for a specific benchmarking case.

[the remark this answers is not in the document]
yflim: On second thought, perhaps [sampling only each DB and transaction entity count combination](https://github.com/replikativ/datahike/pull/496/commits/7ef47a55881f03d7e486980e97afb0b47bbc2136) makes more sense? I changed the code to do that. Do you think the database should still be reinstantiated for each sample? Or that we should still have two separate parameters for DB and transaction entity count?

benchmark/src/benchmark/measure.clj
jsmassa: This should work

benchmark/src/benchmark/measure.clj
jsmassa: There is still a comment section in this namespace, which should be removed before merging.

benchmark/src/benchmark/measure.clj
```diff
@@ -103,23 +128,93 @@
...
-                  {:context context
-                   :time (time-statistics times)}))))
-      
-       (remove nil?)))
+(defn get-measurements
+  ([options] (get-measurements options c/db-configs {}))
+  ([options configs] (get-measurements options configs {}))
+  ([{:keys [db-entity-counts] :as options} configs specified-fn]
+   (->> (for [cfg configs
+              entity-count db-entity-counts]
```
jsmassa, line 136: Should be 
```
([{:keys [db-entity-counts db-samples] :as options} configs specified-fn]
   (->> (for [cfg configs
                  entity-count db-entity-counts
                  _ (range db-samples)]
```

And `db-samples` should be added as cli option. It will make sense to use this for query measurements if the focus is on rare queries that can't benefit from cached values.

jsmassa: So `db-samples` should have the default value `1`.

benchmark/src/benchmark/measure.clj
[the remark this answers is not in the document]
jsmassa: Leave the code on line 73-74 as you have it and the second loop has to got to its original spot, i.e. line 137 now (see comment below), so there will be an actual sampling going on. With that parameter the user will have the choice to either run the benchmarks quickly, by setting _iterations_ to the number of samples wanted and _db-samples_ to 1, which tests a function on the same database _iterations_ times accepting that queries might seem quicker because caches are being used, or the user is interested in query times without caching and so should set _db-samples_ to the number of samples wanted and _iterations_ to 1 which will result in the same number of samples, but it will run slower because the database is built up again for each sample, but it should yield more stable results because it circumvents the caching issue.

src/datahike/migrate.clj
```diff
@@ -1,37 +1,62 @@
...
+          query {:query wd/find-datoms-in-tx
+                 :args [(if (db/-keep-history? db) (api/history db) db)]}]
+      (cbor/spit-all path (mapcat (fn [[tid tinst]]
+                                    (->> (api/q (update-in query [:args] conj tid))
+                                         (sort-by first)
+                                         (into [[tid :db/txInstant tinst tid true]])))
+                                  txs))))
 
+  ;; Dealbreaker: Cannot work with byte arrays
+  (defn export-db-tc
```
kordano, line 55: Since this is clearly not working we can actually remove the code and the dependency.

kordano: So, tablecloth and wanderung.

src/datahike/migrate.clj
```diff
@@ -1,37 +1,62 @@
...
-   (doseq  [datoms (->> (line-seq (io/reader path))
-                        (map read-string)
-                        (partition 1000 1000 nil))]
-     (print ".")
-     (api/transact conn (vec datoms)))))
+  (let [datoms (->> (cbor/slurp-all path)
+                    (map #(-> (apply d/datom %) (update :v instance-to-date))))]
+    (swap! conn update-max-tx datoms)
+    (print "Importing ")
+    (api/transact conn (vec datoms))))
```
kordano, line 32: This means we are importing all datoms at once, so no partitions for now?

test/datahike/test/migrate_test.clj
```diff
@@ -226,3 +219,26 @@
                                :cache-size         0
                                :index :datahike.index/persistent-set
                                :attribute-refs?    false}))
+
+(defn- all-true? [c] (every? true? c))
```
kordano, line 223: Might make sense to move `all-true?` and `all-eq?` to the `datahike.test.utils` namespace at some point, so we could them also in other tests.

kordano (changes requested):

src/datahike/migrate.clj
yflim: Right, how'd I forget

src/datahike/migrate.clj
yflim: I thought to drop them for now since with history, they introduce transaction datoms that don't belong to the imported database. They could be removed, but that seemed clunky, so I figured to drop partitions for now pending a better solution. But either approach makes sense—let me know.

test/datahike/test/migrate_test.clj
yflim: Thought of that but held back. Since you mention it though, why not now? So that they're more visible.

kordano (approved):

Issue #478, opened 2022-02-07
[Bug]: migration with attribute refs causes error


kordano:
### What version of Datahike are you using?

0.4.1480

### What version of Java are you using?

openjdk version "11.0.13" 2021-10-19

### What operating system are you using?

Ubuntu 20.04.3 LTS

### What database EDN configuration are you using?

```edn
{:store              {:backend :mem
                      :id      "export"}
 :keep-history?      true
 :schema-flexibility :write
 :attribute-refs?    true}
```

### Describe the bug

When exporting a database with attribute refs and importing it in a new database the following exception is thrown:
``` clojure
#error {
 :cause "System schema entity cannot be changed"
 :data {:error :transact/schema, :entity-id 1}
 :via
 [{:type clojure.lang.ExceptionInfo
   :message "System schema entity cannot be changed"
   :data {:error :transact/schema, :entity-id 1}
   :at [datahike.db$update_schema invokeStatic "db.cljc" 1426]}]
 :trace
 [[datahike.db$update_schema invokeStatic "db.cljc" 1426]
  [datahike.db$update_schema invoke "db.cljc" 1415]
  [datahike.db$with_datom_upsert invokeStatic "db.cljc" 1592]
  [datahike.db$with_datom_upsert invoke "db.cljc" 1565]
  [clojure.lang.AFn applyToHelper "AFn.java" 156]
  [clojure.lang.AFn applyTo "AFn.java" 144]
  [clojure.core$apply invokeStatic "core.clj" 669]
  [clojure.core$update_in$up__6870 invoke "core.clj" 6174]
  [clojure.core$update_in invokeStatic "core.clj" 6175]
  [clojure.core$update_in doInvoke "core.clj" 6161]
  [clojure.lang.RestFn invoke "RestFn.java" 467]
  [datahike.db$transact_report invokeStatic "db.cljc" 1601]
  [datahike.db$transact_report invoke "db.cljc" 1594]
  [datahike.db$transact_add invokeStatic "db.cljc" 1724]
  [datahike.db$transact_add invoke "db.cljc" 1713]
  [datahike.db$transact_tx_data invokeStatic "db.cljc" 2147]
  [datahike.db$transact_tx_data invoke "db.cljc" 1822]
  [datahike.core$with invokeStatic "core.cljc" 129]
  [datahike.core$with invoke "core.cljc" 122]
  [datahike.core$_transact_BANG_$fn__21030 invoke "core.cljc" 212]
[...]

PR #501, opened 2022-03-22
478 migration with attribute refs causes error

yflim:
#### SUMMARY
Fix exception thrown on importing a database export containing attribute refs #478

##### Bugfix
- [ ] Integration tests added
- [x] Formatting checked
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/src/datahike/migrate.clj
+++ b/src/datahike/migrate.clj
@@ -7,8 +7,9 @@
 
 (defn export-db
   "Export the database in a flat-file of datoms at path."
-  [db path]
-  (let [cfg (:config db)]
+  [conn path]
+  (let [db @conn
+        cfg (:config db)]
     (cbor/spit-all path (cond->> (sort-by
                                   (juxt d/datom-tx :e)
                                   (api/datoms (if (:keep-history? cfg) (api/history db) db) :eavt))
--- a/src/datahike/migrate.clj
+++ b/src/datahike/migrate.clj
@@ -1,5 +1,6 @@
 (ns ^:no-doc datahike.migrate
   (:require [datahike.api :as api]
+            [datahike.constants :as c]
             [datahike.datom :as d]
             [datahike.db :as db]
             [clj-cbor.core :as cbor]))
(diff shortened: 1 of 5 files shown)


yflim:
Note that most of the commits here are from branch [477-history-migration-not-working](https://github.com/replikativ/datahike/tree/477-history-migration-not-working) for [PR 496](https://github.com/replikativ/datahike/pull/496). The only one strictly for this issue/PR is the last, [743db12](https://github.com/replikativ/datahike/commit/743db12fffd6f529967fc558b9b0ae4ad1f7a198).

yflim:
Not sure what "This can happen as long as we are arriving at version 1.0.0 and will always be communicated through the Changelog" under [Migration & Backup](This can happen as long as we are arriving at version 1.0.0 and will always be communicated through the Changelog) in the README means, so left it alone even if it should be edited.

jsmassa:
Should be rebased after #496 has been merged to make actual changes more visible

kordano (changes requested):

src/datahike/migrate.clj
```diff
@@ -1,15 +1,20 @@
...
-  (cbor/spit-all path (map seq (sort-by
-                                (juxt d/datom-tx :e)
-                                (api/datoms (if (db/-keep-history? db) (api/history db) db) :eavt)))))
+  [conn path]
+  (let [db @conn
+        cfg (:config db)]
+    (cbor/spit-all path (cond->> (sort-by
+                                  (juxt d/datom-tx :e)
+                                  (api/datoms (if (:keep-history? cfg) (api/history db) db) :eavt))
+                          (:attribute-refs? cfg) (remove #(= (d/datom-tx %) c/tx0))
```
jsmassa, line 16: What happens if the system entities are changed between export and import datahike versions? For example when the order changes or there are more or less system entities than before?

yflim: I assumed any changes to system entities would only be additive. I guess that isn't a sufficiently realistic or safe assumption? Also, as I understand, the migration code here isn't a permanent solution—a role Wanderung is supposed to fulfil at some point. Please let me know.

jsmassa: I guess, if it is only a temporary solution, it should be ok for now, but maybe put at least a note in there about the assumption. 
I was of the impression, the system entities are in the database for exactly that reason - so the case of invalid references would never occur. But I wouldn't be surprised if none of our functions support importing the schema and I'd say that would be my mistake, sorry. I will open an issue or discussion about it to not lose track of the subject.

kordano (approved):

PR #503, opened 2022-03-28
Durable persistent set

jsmassa:
#### SUMMARY
Enable all stores to use with persistent set

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [x] Formatting checked
- [x] `CHANGELOG.md` updated


src/datahike/index/persistent_set.cljc
```diff
@@ -31,54 +24,103 @@
...
+    pset)
+  (-transient [pset]
+    (transient pset))
+  (-persistent! [pset]
+    (persistent! pset)))
+
+(defn get-loader [konserve-store]
+  (proxy [Loader] []
+    (load [address]
+      (let [children-as-maps (async/<!! (k/get konserve-store address))]
```
whilo, line 75: It is not good to use `<!!` here, it would be better to use synchronous konserve calls instead.

src/datahike/index/persistent_set.cljc
```diff
@@ -31,54 +24,103 @@
...
+  (assoc store :serializers {:FressianSerializer (fressian-serializer
+                                                  {"datahike.index.PersistentSortedSet" (reify ReadHandler
+                                                                                          (read [_ reader _tag _component-count]
+                                                                                            (let [{:keys [index-type root-node]} (.readObject reader)
+                                                                                                  loader (get-loader store)
+                                                                                                  pset (with-meta (set/sorted-set-by (index-type->cmp index-type) loader)
+                                                                                                         {:index-type index-type})]
+                                                                                              (set! (._root pset)
+                                                                                                    (set/map->node loader root-node))
```
whilo, line 107: It would be better to provide serializers for the Node/Leaf class of the persistent sorted set itself, then we would not have to (de)serialize them in the loader, but can instead do it transparently in the storage layer.

whilo (commented):
Nice work. I assume that we want to ship this asap and then tell people to do manual migrations when we change the persistent-sorted-set? Otherwise we also would have to cover versioning in the persistent sorted set to be able to automatically upgrade later.

src/datahike/index/persistent_set.cljc
jsmassa: As far as I know, this is not enabled yet for datahike. I added a note to change this later though.

jsmassa:
Fixes #519

jsmassa:
Fixes #507 by enabling identical tests

src/datahike/index/persistent_set.cljc
```diff
@@ -1,93 +1,253 @@
...
+        old-datoms (cond->> (slice pset
+                                   (dd/datom (.-e datom) (.-a datom) nil tx0)
+                                   (dd/datom (.-e datom) (.-a datom) nil txmax)
+                                   index-type)
+                     (= :avet index-type)
+                     (filter #(= (.-e datom) (.-e %))))
+        {added-datoms true
+         retracted-datoms false} (group-by :added old-datoms)
+        old-val (first (clojure.set/difference (set (map :v added-datoms))
+                                               (map :v retracted-datoms)))]
```
yflim, line 98: This doesn't work if an old value is re-upserted and then replaced, does it? E.g. upsert value `a`, `b`, then `a` again, followed by `c`—if I'm not mistaken, `a` would be added again but not retracted when `c` is upserted.

src/datahike/index/persistent_set.cljc
whilo: This is not critical though, in particular if we do manual migrations for now anyway.

src/datahike/index/persistent_set.cljc
whilo: This set logic definitely would not retain order as far as I can see.

whilo: This had multiple problems, most importantly it is pulling all historical data to do set operations on the full history, which scales in cost with the size of the history for each `[e a]` pair. Instead we should just retrieve the old-value from `eavt` and pass it in from the outside (in general we need to use indices wisely whenever possible). We could not do this with the hitchhiker-tree, because we wanted to avoid lookups and be lazy with writes, but DataScript does the same thing for upserts and the persistent-sorted-set is more than fast enough for that. This change renders upserts in the transaction benchmark sandbox as fast as inserts for me.

whilo: Note also that this has nothing to do with durability, it also improves the persistent-sorted-set memory speed.

whilo:
Update: Thread safety is now covered in an update to the persistent-sorted-set, providing (hopefully) safe bounded memory usage under concurrent reads (and cache evictions). I have stress tested the implementation with the integrated benchmark suite and small cache size, but there is still an issue with store management where I think the store is deleted including a needed value before the query tests are run. This potentially was not visible before as more data was hold in memory. This issue needs to be solved and it needs to be verified that performance satisfies the requirements in general. We also need to decide where to put the cache and make sure we only have one per database. Other than that I think we are mostly there.

whilo:
Ok, I have mostly completed the open issues I saw and see this PR as basically done. Note that cache sizes now need to be greater than 0 because the cache constructors do not support 0 sized caches (there was still a `1000` hardcoded, I think accidentally). I can do further performance tuning of the comparators and the set if needed, otherwise I would suggest to do this in another PR. I think we want to have a persistent-sorted-set release before merging this, I can coordinate with @tonsky to land this.

jsmassa:
I have worked on the upsert comparator to get the migration performance back to a sensible level. The jobtech benchmarks are also showing good results now, so from this point of view, this PR is finally ready now. 

A release of the set before merging would be nice.

The use of the `:cache-size` parameter for konserve confused me. It should be an separate parameter, so the caches can be controlled independent of each other. I will change it accordingly.

whilo:
I agree, but I would also say that ideally we should only need the persistent-sorted-set cache now. We should deprecate and drop the hitchhiker-tree as backend once the set performs sufficiently well and then its caching mechanism is providing the same logic as the underlying konserve cache had, so that will be redundant and I think we can drop the konserve cache. 

I think the same should hold for the cache we currently have for slicing (search), the persistent sorted set should be fast enough to drop it, unless our comparators slow down the slicing so much that we need to cache the results because of this slowness. In terms of pure iterator speed on top of integers the pss is the same for me as Clojure's persistent vector which is used for the slicing cache. The benefit of focusing all the caching budget we have now to the pss cache is that it is shared between all database instances even while transacting into the databases while also deduplicating any overlap that would exist between different slices of the indices cached. We have to measure the impact of this though.

src/datahike/connector.cljc
```diff
@@ -23,13 +23,14 @@
         ;; flush for in-memory backends would only make sense if multiple processes access the db
         ;; TODO: update can also be skipped for external stores when single process writes unless memory should be freed up
         ;;       -> add options to config
-        eavt-flushed (cond-> eavt (not= :mem (-> config :store :backend)) (di/-flush backend))
-        aevt-flushed (cond-> aevt (not= :mem (-> config :store :backend)) (di/-flush backend))
-        avet-flushed (cond-> avet (not= :mem (-> config :store :backend)) (di/-flush backend))
+        backend? (not= :mem (-> config :store :backend))
```
whilo, line 26: I would suggest to push this logic into `di/-flush` instead of manually dispatching around its callsites.

jsmassa: If it's important to you, you are free to change it accordingly.

jsmassa:
We can set the default parameters accordingly to promote the parameter combination we see as optimal. The search cache can be turned off completely by setting it to 0 anyway. I wouldn't drop anything (yet) though since at some point we also want to explore more (elaborate) data structures for the index and eventually have one that is more engineered to the problem and integrated into the project. Until then it's nice to have some options for experimentation.

src/datahike/db.cljc
```diff
@@ -1553,26 +1553,25 @@
         keep-history? (and (-keep-history? db) (not (no-history? db a-ident))
                            (not= :db/txInstant a-ident))
         op-count      (.op-count db)
-        old-val (when keep-history?
-                  (:v (first (di/-slice (:eavt db)
-                                        (dd/datom (.-e datom) (.-a datom) nil (.-tx datom))
-                                        (dd/datom (.-e datom) (.-a datom) nil (.-tx datom))
-                                        :eavt))))]
+        old-datom (first (di/-slice (:eavt db)
```
jsmassa, line 1556: Sorry, but just like `old-val` this doesn't belong in this place. As seen in the hitchhiker-tree, each index type has to decide for itself how to handle upserts, otherwise the index upsert function doesn't even make sense anymore. I will move this logic back into the index namespace. It is correct though that we can simply use the `:eavt` index to retrieve the old datom. I can try out if that is enough to have a decent transaction performance but I doubt it.

jsmassa:
I see, I also have to clean up some clutter I brought in.

src/datahike/db.cljc
jsmassa, line 1556: This seems to be much slower than as I had it, at least in the jobtech tests. It also makes having an upsert function unnecessary. So like `old-val` retrieving the `old-datom` should be done in the index namespace.

jsmassa: For some reason this also breaks the benchmarks I am trying to run, although that doesn't seem to make much sense to me.

jsmassa:
I started to investigate the current problems and found out that both the current version and the prior version with my changes are not working properly with the jobtech benchmarks. 
@kordano has encountered this problem as well and is working on finding the root cause.

jsmassa:
Blocked by #542

whilo (approved):
Looks good to me.

jsmassa:
I have run the jobtech benchmarks with 100 iterations and the results look fine. The median of the measurements is almost the same and the average is usually better.
Here are some measurements (in ms):

base-urls:
|          main          |            dps       |   
| ---------------------- | ---------------------- |
|                29.028 |                29.436 |
|                52.818 |                90.214 |
|            2460.009 |           2100.082 |

 graphql-basic:
|          main          |            dps       |   
| ---------------------- | ---------------------- |
|              7416.746 |              3491.175 |
|                27.092 |                35.937 |
|               961.287 |               455.830 |

graphql-benchmark:
|   main   |          |   dps   |          |
| -------- | ----------- | -------- | ----------- |
|   min   |     max     |   min   |     max     |
|   167.5 |      2091.6 |   168.1 |      1575.4 |
|    23.8 |    195350.6 |    24.1 |     91244.2 |
|    10.2 |       952.2 |     9.5 |       448.0 |
|     9.3 |       669.6 |     9.4 |       334.0 |
|    17.3 |     12481.2 |    18.8 |      5693.7 |
|     8.9 |       946.2 |     9.0 |       450.0 |

 endpoints-benchmark:
|   main   |          |   dps   |          |
| -------- | ----------- | --------- | ---------- |
|   min   |     max     |   min   |     max     |
|     9.1 |        14.0 |     8.9 |        14.6 |

In other runs the base-url benchmark and the graphql-basic benchmarks were better with this branch than with the main branch, so all-in-all I would suggest finally merging this into main.

What do you think, @kordano?

kordano (commented):
Looks good, but fails on my machine with:
```
Syntax error compiling at (datahike/store.cljc:101:5).
No such var: fs/connect-fs-store
```

deps.edn
```diff
@@ -4,7 +4,7 @@
         io.replikativ/hitchhiker-tree               {:mvn/version "0.2.222"}
         io.replikativ/incognito                     {:mvn/version "0.3.66"}
         io.replikativ/konserve                      {:mvn/version "0.7.275"}
-        persistent-sorted-set/persistent-sorted-set {:mvn/version "0.1.4"}
+        persistent-sorted-set/persistent-sorted-set {:local/root "persistent-sorted-set-0.1.4.jar"} ;; TODO: eventually clojars object
```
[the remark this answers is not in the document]
jsmassa: Good question. I don't know much about how the packaging works. Is there a way to include the jar for now? Otherwise we would need a proper release for our [durability branch of the persistent set](https://github.com/replikativ/persistent-sorted-set/tree/durability-with-pom).

Ideally, having our own fork of the set including the durability layer is only a temporal solution until Nikita has included an implementation in his library. He is currently working it on a [branch of the original persistent set](https://github.com/tonsky/persistent-sorted-set/tree/storage).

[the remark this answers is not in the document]
whilo: This is fixed now. As I had mentioned, Nikita already released the persistent-sorted-set a week ago and we can just use his upstream version for now.

whilo:
> Looks good, but fails on my machine with:
> 
> ```
> Syntax error compiling at (datahike/store.cljc:101:5).
> No such var: fs/connect-fs-store
> ```

This indicates that you use an old konserve version somehow.

whilo (approved):
Good job :smile: !

mpenet:
nice one indeed!

whilo:
@mpenet Nice to hear from you! https://github.com/replikativ/datahike/pull/232 is about to be merged next and should provide some of the functionality you were interested in (gc and versioning), let me know what you think and whether anything could be added to make it work better for you.

PR #506, opened 2022-04-04
Add comment that datahike.migrate is meant to be a temporary solution

yflim:
#### SUMMARY
Add comment [suggested](https://github.com/replikativ/datahike/pull/501#discussion_r841258863) in discussion on PR 501 (migration with attribute refs causes error).

### DIFF
--- a/src/datahike/migrate.clj
+++ b/src/datahike/migrate.clj
@@ -5,6 +5,8 @@
             [datahike.db :as db]
             [clj-cbor.core :as cbor]))
 
+;; Note: the code here is intended to be a temporary solution, pending developments in Wanderung
+
 (defn export-db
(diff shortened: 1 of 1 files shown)


src/datahike/migrate.clj
```diff
@@ -5,6 +5,8 @@
             [datahike.db :as db]
             [clj-cbor.core :as cbor]))
 
+;; Note: the code here is intended to be a temporary solution, pending developments in Wanderung
```
kordano, line 8: This should go into the docs for both functions.

kordano (changes requested):

kordano (approved):
thanks @yflim

PR #509, opened 2022-04-06
Fix typo in comparison predicate in benchmark query

yflim:
Based on the names and contents of the benchmarking functions `less-than-query-1-fixed` and `equals-query-1-fixed`, it seems `=` was used where `<` was intended in the former.


jsmassa (approved):
You are correct, that looks like a copying error

Issue #489, opened 2022-02-21
[Bug]: Benchmarks aren't working


Ramblurr:
### What version of Datahike are you using?

tags/0.4.1484

### What version of Java are you using?

openjdk 11.0.14 2022-01-18

### What operating system are you using?

Fedora Linux

### What database EDN configuration are you using?

file and mem

### Describe the bug

The benchmarks are crashing out of the box.

### What is the expected behaviour?

It crashes.

### How can the behaviour be reproduced?

1. Clone the repo (I'm using `main` which as of time of report is `97eb463b8c71c60f9ff261db795ee679e4e78767`)
2. Copy and paste the [provided benchmark examples](https://github.com/replikativ/datahike/blob/main/doc/benchmarking.md#examples)
3. Observe that it crashes

All 3 provided examples crash with similar logs. For example:

```
$ TIMBRE_LEVEL=':warn'  clj -M:benchmark run -f :query -q :simple-query -i 10
Compile-time (elision) Timbre min-level: :warn
Syntax error (ClassNotFoundException) compiling at (datahike/impl/entity.cljc:1:1).
datahike.java.IEntity

Full report at:
/tmp/clojure-10710237421126008452.edn
```
full report: https://gist.github.com/Ramblurr/34eef3edbc3cb0d7735303e1e1a080b4

kordano:
Could you check what happens if you run `clj -T:build compile` and then run the benchmarks again? Looks like somehow the java bindings are used in here but I don't understand why it is used in the benchmark. Maybe @jsmassa has also some ideas on what might go wrong here.

QuinnWilton:
Hi @kordano! I had the same issue as the issue reporter, and can confirm that running `clj -T:build compile` beforehand fixes the benchmarks.

Thanks!

Ramblurr:
Also can confirm that fixes it. Maybe just add a small note to the docs? Compiling beforehand is painless enough.

```console
$ clj -T:build compile
...
$ TIMBRE_LEVEL=':warn'  clj -M:benchmark run -f :query -q :simple-query -i 10

...
```

TimoKramer:
I am assigning this to me because there should be [this](https://clojure.org/reference/deps_and_cli#prep) in the deps.edn.

I will also improve the docs.

PR #511, opened 2022-04-09
docs: improve docs and add prep step

TimoKramer:
#### SUMMARY
Improve docs as mentioned in #489 to make it clearer that there needs to be a compilation for working on the source code. This is something a lot of people are tripping over when starting to work on our source. I also added a prepare-step to our deps.edn like it is described here: https://clojure.org/reference/deps_and_cli#prep. 

##### Docs
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## next minor/major release
 
+- Improve docs
+- Add prep-step to deps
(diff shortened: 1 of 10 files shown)


kordano (approved):
Looks good, thanks for integrating all these small issues.

PR #513, opened 2022-04-12
Add migration tests

jsmassa:
Just 2 tests added to test config compatibility for migrations. Also tried attribute references to/from keyword databases but that wouldn't work, so I removed them. Would also be harder to implement here, but will be provided with wanderung


kordano (approved):
Thanks for adding the test :)

PR #515, opened 2022-04-21
Datom and index metrics

yflim:
#### SUMMARY
Add datom and index metrics

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [ ] Formatting checked
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -11,8 +11,9 @@
    [datahike.db.utils :as dbu]
    [datahike.index :as di]
    [datahike.schema :as ds]
-   [taoensso.timbre :refer [warn]]
-   [datahike.tools :as tools :refer [raise]])
+   [datahike.tools :as tools :refer [raise]]
+   [medley.core :as m]
+   [taoensso.timbre :refer [warn]])
   #?(:cljs (:require-macros [datahike.db :refer [defrecord-updatable]]
                             [datahike.datom :refer [combine-cmp datom]]
                             [datahike.tools :refer [case-tree raise]]))
@@ -71,15 +72,15 @@
      (let [impl-map (->> impls (map (juxt get-sig identity)) (filter first) (into {}))
            body (macroexpand-1 (list* 'defrecord name fields impls))]
        (postwalk
-        (fn [form]
-          (if (and (sequential? form) (= 'deftype* (first form)))
-            (->> form
-                 dedupe-interfaces
-                 (remove (fn [method]
-                           (when-let [impl (-> method get-sig impl-map)]
-                             (not= method impl)))))
-            form))
-        body))))
+         (fn [form]
+           (if (and (sequential? form) (= 'deftype* (first form)))
+             (->> form
+                  dedupe-interfaces
+                  (remove (fn [method]
+                            (when-let [impl (-> method get-sig impl-map)]
+                              (not= method impl)))))
+             form))
+         body))))
 
 #?(:clj
    (defn- make-record-updatable-cljs [name fields & impls]
@@ -90,8 +91,8 @@
 #?(:clj
    (defmacro defrecord-updatable [name fields & impls]
      `(if-cljs
-       ~(apply make-record-updatable-cljs name fields impls)
-       ~(apply make-record-updatable-clj name fields impls))))
+        ~(apply make-record-updatable-cljs name fields impls)
+        ~(apply make-record-updatable-clj name fields impls))))
 
 ;; TxReport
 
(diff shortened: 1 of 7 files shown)


yflim:
Does it make sense to provide count and index metrics on database records besides DB (HistoricalDB, FilteredDB at the moment)? Thought it could be useful, hence -metrics is in the IDB protocol for now so it can be implemented across database records.

src/datahike/db.cljc
```diff
@@ -1035,6 +1035,25 @@
...
+     :per-entity-counts (reduce (fn [oecs aecs]
+                                  (reduce-kv
+                                   (fn [ecs e c] (if (ecs e) (update ecs e #(+ % c)) (assoc ecs e c)))
+                                   oecs aecs))
+                                {}
+                                (map #(into {} (map (fn [[e d]] [e (count d)]) %)) (vals attr-entity-counts)))
+     :avet-count (reduce + 0 (map second (filter-idxed-attrs per-attr-counts)))
+     :temporal-count (-count (.-temporal-eavt db))
+     :temporal-avet-count
+     (reduce + 0 (map count (vals (filter-idxed-attrs (group-by :a (-seq (.-temporal-eavt db)))))))}))
```
jsmassa, line 1055: Use `->>`

src/datahike/db.cljc
```diff
@@ -1035,6 +1035,25 @@
...
+  {:pre [(instance? DB db)]}
+  (let [attr-entity-counts (into {} (map (fn [[a d]] [a (group-by :e d)]) (group-by :a (-seq (.-eavt db)))))
+        per-attr-counts
+        (into {} (map (fn [[a ent-ds]] [a (reduce + 0 (map count (vals ent-ds)))]) attr-entity-counts))
+        filter-idxed-attrs (fn [attr-kvs] (filter (fn [[k v]] (-> (.-rschema db) :db/index k)) attr-kvs))]
+    {:count (-count (.-eavt db))
+     :per-attr-counts per-attr-counts
+     :per-entity-counts (reduce (fn [oecs aecs]
+                                  (reduce-kv
+                                   (fn [ecs e c] (if (ecs e) (update ecs e #(+ % c)) (assoc ecs e c)))
```
jsmassa, line 1048: `if` statement should have if and else branches in lines on their own

src/datahike/db.cljc
```diff
@@ -1035,6 +1035,25 @@
...
+                                   oecs aecs))
+                                {}
+                                (map #(into {} (map (fn [[e d]] [e (count d)]) %)) (vals attr-entity-counts)))
+     :avet-count (reduce + 0 (map second (filter-idxed-attrs per-attr-counts)))
```
jsmassa, line 1052: Potentially use `->>`

src/datahike/db.cljc
```diff
@@ -1035,6 +1035,25 @@
...
+        (into {} (map (fn [[a ent-ds]] [a (reduce + 0 (map count (vals ent-ds)))]) attr-entity-counts))
+        filter-idxed-attrs (fn [attr-kvs] (filter (fn [[k v]] (-> (.-rschema db) :db/index k)) attr-kvs))]
+    {:count (-count (.-eavt db))
+     :per-attr-counts per-attr-counts
+     :per-entity-counts (reduce (fn [oecs aecs]
+                                  (reduce-kv
+                                   (fn [ecs e c] (if (ecs e) (update ecs e #(+ % c)) (assoc ecs e c)))
+                                   oecs aecs))
+                                {}
+                                (map #(into {} (map (fn [[e d]] [e (count d)]) %)) (vals attr-entity-counts)))
```
jsmassa, line 1051: Hard to read: Which `map` has `(vals attr-entity-counts)` as parameter? The first or the second one? It's not obvious with just one look. Solution: threading macro or separate line for collection.

jsmassa (changes requested):
Lacks readability. 
I would really encourage you to use more [threading macros](https://clojure.org/guides/threading_macros) and not to hesitate to put expressions on a new line when things get too nested.
Also, even if it is just to keep some order, within a hash-map, please, stick to either putting the value next to its key or below it, but don't mix it up!

yflim:
> Lacks readability. I would really encourage you to use more [threading macros](https://clojure.org/guides/threading_macros) and not to hesitate to put expressions on a new line when things get too nested. Also, even if it is just to keep some order, within a hash-map, please, stick to either putting the value next to its key or below it, but don't mix it up!

Then again, why do all that instead of simply `(group-by :a)` and `(group-by :e)` on `(-seq (.-eavt db))` (invoked here anyway) and its temporal equivalent? This buys neither memory nor performance, I think. Am I missing or forgetting anything?

jsmassa:
What do you mean? Do you want to somehow include it in the `-seq` function or count everything stepping through the sequence just once?

yflim:
I mean something like (abridged for clarity):
`(let [eavt-datoms (-seq (.-eavt db))]
    {:per-attr-counts (->> (vals (group-by :a eavt-datoms))
                                           (map count))
     :per-entity-counts (->> (vals (group-by :e eavt-datoms))
                                              (map count))})`

jsmassa:
Sure, you can write it like that. I don't think it has performance benefits either though. I would expect that it will still step through the sequence twice; once for each of the grouping operations. To only step through it once, you could implement a big reduce operation with the empty count-map as initial value and then set/increase the counts with `assoc-in` or `update-in` operations.

It definitely makes sense to implement one function and then call it for the current and for the temporal indices. 

I also think, this implementation will not result in what you actually wanted to get. By calling `vals`, you lose the information of which attribute the count corresponds to. I think, what you would want is something like:
```
(->> eavt-datoms
     (group-by :e)
     (map-vals count))
```
(The `map-vals` function is not part of the clojure core library, but can be pulled in with the [medley](http://weavejester.github.io/medley/medley.core.html#var-map-vals) library or has to be written yourself.)

Btw, note how I am using even more lines of code here for the threading macro. This is the easiest to read: start with the object you want to transform on the first line and then list each transformation from top to bottom each on their own line even if it's something short like `vals`.

src/datahike/db.cljc
```diff
@@ -1035,6 +1035,33 @@
...
+                             (-> (update-count m [:per-attr-counts (.-ident-for db (.-a d))] update-in)
+                                 (update-count [:per-entity-counts (.-e d)] update-in)))
+                           {:per-attr-counts    {}
+                            :per-entity-counts  {}}
+                           eavt-datoms)
+        temp-per-attr-counts (when (-keep-history? db)
+                               (reduce (fn [m d] (update-count m (.-ident-for db (.-a d)) update))
+                                       {}
+                                       (-seq (.-temporal-eavt db))))
+        filter-idxed-attrs (fn [attr-counts] (filter (fn [[a v]] (-> (:db/index (.-rschema db))
```
jsmassa, line 1052: Why is this a separate function? Isn't it just needed once?

jsmassa (dismissed):

jsmassa (approved):

Issue #507, opened 2022-04-06
[Bug]: 


jsmassa:
### What version of Datahike are you using?

0.4.1489

### What version of Java are you using?

openjdk 11.0.13 2021-10-19

### What operating system are you using?

Ubuntu

### What database EDN configuration are you using?

Any

### Describe the bug

The [duplicate-test](https://github.com/replikativ/datahike/blob/85597f45599a2f84345dd2a14fb7555d6fcb9281/test/datahike/test/insert.cljc#L13)) indicates that there is still a difference for "identical" datoms when added to the database. 
In the hitchhiker tree a datom doesn't get overwritten when another one with the same eav values comes in, whereas in the persistent set this might still not be the case. 
This should be checked for identical behavior to avoid problems in the future. If it already shows the correct behavior the test/comment should be adjusted.

### What is the expected behaviour?

Same behavior for hitchhiker-tree index and persistent set index

### How can the behaviour be reproduced?

Run/adjust the [duplicate-test](https://github.com/replikativ/datahike/blob/85597f45599a2f84345dd2a14fb7555d6fcb9281/test/datahike/test/insert.cljc#L13) for the persistent set index.

PR #516, opened 2022-04-22
Consistent insert behaviour for hitchhiker-tree and persistent sorted set

yflim:
#### SUMMARY
Fix persistent sorted set behaviour on insert to be consistent with hitchhiker-tree behaviour, so that a datom doesn't get overwritten when another one with the same EAV values is inserted. #507 

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/src/datahike/index/persistent_set.cljc
+++ b/src/datahike/index/persistent_set.cljc
@@ -56,12 +56,11 @@
   (set/disj set datom (index-type->cmp-quick index-type)))
 
 (defn -insert [set datom index-type]
-  (-> (or (when-let [old (first (-slice set
-                                        (dd/datom (.-e datom) (.-a datom) (.-v datom) tx0)
-                                        (dd/datom (.-e datom) (.-a datom) (.-v datom) txmax)))]
-            (-remove set old index-type))
-          set)
-      (set/conj datom (index-type->cmp-quick index-type))))
+  (if (-slice set
+              (dd/datom (.-e datom) (.-a datom) (.-v datom) tx0)
+              (dd/datom (.-e datom) (.-a datom) (.-v datom) txmax))
+    set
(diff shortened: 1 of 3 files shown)


src/datahike/index/persistent_set.cljc
```diff
@@ -56,12 +56,11 @@
...
-                                        (dd/datom (.-e datom) (.-a datom) (.-v datom) tx0)
-                                        (dd/datom (.-e datom) (.-a datom) (.-v datom) txmax)))]
-            (-remove set old index-type))
-          set)
-      (set/conj datom (index-type->cmp-quick index-type))))
+  (if (-slice set
+              (dd/datom (.-e datom) (.-a datom) (.-v datom) tx0)
+              (dd/datom (.-e datom) (.-a datom) (.-v datom) txmax))
+    set
+    (set/conj set datom (index-type->cmp-quick index-type))))
```
jsmassa, line 63: Looks correct to me. Can you also remove the `-insert` and `-remove` duplicates (l. 34/37) though? I think they got duplicated a while ago in the course of a merge or rebase.

jsmassa (dismissed):
Looks good. Thank you for the fix :)

yflim:
This probably shouldn't be merged until (unless?) the inconsistent behaviour is also fixed for upsert though. What do you think?

yflim:
> I just saw that there is also an entry missing in the changelog about this. Maybe something like 'Insert into persistent set does not replace existing datom', so something that describes it well what actually has been changed.

Right, will do.

jsmassa:
I think those issues can be separate PRs. I would like to merge the upsert change this week as well though since that stuff is a blocker now for the durable persistent set.

jsmassa (approved):

PR #518, opened 2022-04-22
ci: retry release with backoff

TimoKramer:


jsmassa (approved):
Looks good to me. Thanks for the fix :) Hopefully this will do the trick

Issue #524, opened 2022-04-27
Datom comparator refactoring


jsmassa:
### Describe the feature you would like to request

The hitchhiker-tree index and the persistent-set index use different datom comparators. Ideally, they would use the same ones to reduce maintenance efforts. 

Additionally, the treatment of retract attributes (`added? false`) should be thought about again. So far, in the hitchhiker-tree, the `added?` flag flips the sign of the transaction ID, which messes up the order and with that stops other optimizations to be implemented. Besides, in non-historical data, this flag is not needed.
 

### Describe the solution you would like

1) Find efficient datom comparators and use them in all index types
2) Consider using different comparators for current index and historical index

### Describe alternatives you've considered

- leave everything as is and live with the consequences
- remove the hitchhiker-tree index entirely

PR #526, opened 2022-04-29
Consistent upsert for hitchhiker-tree and persistent-sorted-set indexes

yflim:
#### SUMMARY
Upserting datom with identical EAV to an existing one leaves hitchhiker-tree temporal indices unchanged, but adds to persistent-sorted-set temporal indices.

As for upserting a datom matching an existing EA but with a new V, the old V is not negated; rather, the new datom is merely added, so that the entire history for that EA consists only of additions.

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked
- [ ] `CHANGELOG.md` updated

#### ADDITIONAL INFORMATION
The reasons for this:
1. `-slice` always returns zero matches.
2. The comparator used ignores additions and negations, so a negation of an existing datom evaluates as equal and isn't added to the temporal index.

Added a `datahike.index.utils` namespace with helper functions.

### DIFF
--- a/src/datahike/datom.cljc
+++ b/src/datahike/datom.cljc
@@ -218,26 +218,47 @@
      :clj
      (.compareTo ^Comparable a1 a2)))
 
-(defn cmp-datoms-eavt-quick [^Datom d1, ^Datom d2]
-  (combine-cmp
-   (#?(:clj Long/compare :cljs -) (.-e d1) (.-e d2))
-   (cmp-attr-quick (.-a d1) (.-a d2))
-   (compare (.-v d1) (.-v d2))
-   (#?(:clj Long/compare :cljs -) (datom-tx d1) (datom-tx d2))))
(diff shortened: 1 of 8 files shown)


yflim:
Note: the first commit on this branch is from [fix-hht-slice-early-termination](https://github.com/replikativ/datahike/tree/fix-hht-slice-early-termination); it contains changes necessary to the ones here.

jsmassa:
I ran some benchmarks on the branch and compared it to main. It seems like this PR will make the performance worse for joins on the attribute, but improve the performance for the query named `:less-than-query-1-fixed` run with integer values. Just something to keep in mind. 
The comparators should be revised anyway at some point as stated in #524. If you have any more thoughts about that topic, you are welcome to add them in that issue.

jsmassa (dismissed):
Thanks for the fix!

kordano (approved):

Issue #527, opened 2022-05-01
[Bug]: Adding a lot of entities in individual transactions when using the `:file` backend eventually throws an exception.


phronmophobic:
### What version of Datahike are you using?

0.5.1500

### What version of Java are you using?

openjdk version "17" 2021-09-14 OpenJDK Runtime Environment Temurin-17+35 (build 17+35) OpenJDK 64-Bit Server VM Temurin-17+35 (build 17+35, mixed mode)

### What operating system are you using?

MacOS 12.3.1

### What database EDN configuration are you using?

```clojure
{:store {:backend :file :path (.getAbsolutePath
                               (io/file "my-db"))}}
```

### Describe the bug

Adding a lot of entities in individual transactions when using the `:file` backend eventually throws an exception. The exception does not occur when using the `:mem` backend. The exception also did not occur when adding the same entities all in one large transaction.

Below is the logged error.

```clojure
Supervisor: #error {
 :cause No implementation of method: :-affects-key of protocol: #'hitchhiker.tr
ee.op/IOperation found for class: incognito.base.IncognitoTaggedLiteral
 :via
 [{:type java.lang.IllegalArgumentException
   :message No implementation of method: :-affects-key of protocol: #'hitchhike
r.tree.op/IOperation found for class: incognito.base.IncognitoTaggedLiteral
   :at [clojure.core$_cache_protocol_fn invokeStatic core_deftype.clj 583]}]
 :trace
 [[clojure.core$_cache_protocol_fn invokeStatic core_deftype.clj 583]
  [clojure.core$_cache_protocol_fn invoke core_deftype.clj 575]
  [hitchhiker.tree.op$eval10199$fn__10200$G__10186__10205 invoke op.cljc 8]
  [clojure.core$sort_by$fn__5994 invoke core.clj 3119]
  [clojure.lang.AFunction compare AFunction.java 51]
  [java.util.TimSort countRunAndMakeAscending TimSort.java 355]
  [java.util.TimSort sort TimSort.java 234]
  [java.util.Arrays sort Arrays.java 1233]
  [clojure.core$sort invokeStatic core.clj 3103]
  [clojure.core$sort_by invokeStatic core.clj 3107]
  [clojure.core$sort_by invoke core.clj 3107]
  [hitchhiker.tree.messaging$enqueue invokeStatic messaging.cljc 58]
  [hitchhiker.tree.messaging$enqueue invoke messaging.cljc 31]
[...]

PR #529, opened 2022-05-03
Update read handlers for index insert and upsert ops.

phronmophobic:
#### SUMMARY

Addresses https://github.com/replikativ/datahike/issues/527. I'm not familiar enough with datahike to know if this is a complete fix. Although I was able to reproduce the issue reliably, I was never able to figure out exactly which circumstances trigger the event.

The read handlers for `datahike.index.hitchhiker_tree.insert.temporal_InsertOp` and `'datahike.index.hitchhiker_tree.upsert.temporal_UpsertOp` incorrectly had a `-` instead of `_` which meant that the default read handler was used instead.


jsmassa (approved):
Thanks for the fix, we really appreciate it :)

phronmophobic:
Datahike is really cool! Thank you!

PR #535, opened 2022-05-09
add backend and lib versions to database metadata

yflim:
#### SUMMARY
Add backend and lib versions to database metadata

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [ ] Formatting checked
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -970,7 +970,7 @@
         :system-entities system-entities
         :ref-ident-map ref-ident-map
         :ident-ref-map ident-ref-map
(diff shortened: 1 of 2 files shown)


jsmassa:
So, yes, backend specific information should be treated by the respective backend library, so datahike doesn't have to pull potentially huge additional libraries.

whilo:
I think in case you want to have this metadata in there but not pull dependencies there should be a metadata protocol for the Datahike backends and you just invoke this in `meta-data`. I do not know the background of this PR though.

PR #536, opened 2022-05-12
Fix core.async deadlock in send-transaction!

sundbry:
send-transaction! was blocking (go) threads on the parking (>!!)
operation.

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked
- [ ] `CHANGELOG.md` updated

### DIFF
--- a/src/datahike/transactor.cljc
+++ b/src/datahike/transactor.cljc
@@ -1,7 +1,7 @@
 (ns ^:no-doc datahike.transactor
   (:require [superv.async :refer [<??- S thread-try]]
             [taoensso.timbre :as log]
-            [clojure.core.async :refer [>!! chan close! promise-chan put!]])
+            [clojure.core.async :refer [chan close! promise-chan put!]])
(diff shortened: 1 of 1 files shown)


TimoKramer (commented):
Thanks for that @sundbry. Sorry for the delay. Very much appreciated.

TimoKramer (approved):

whilo:
This is definitely better than potential deadlocks, but it does not exert backpressure. I have converted all transactor calls to non-blocking core.async calls here, /blob/c12d5902d85f0159488103f55e5e04b3e1620293/src/datahike/transactor.cljc, which is better. Feel free to borrow as much from the this PR as needed and I will rebase it.

sundbry:
Hi @whilo, I agree we should figure out some way to exert backpressure here. I will have to come back to this. It looks like that commit is part of this branch https://github.com/replikativ/datahike/pull/337/files, there is a lot going on there, it is hard to see what the specific changes were in re: send-transaction backpressure.

PR #538, opened 2022-05-19
Extend benchmark tuning capabilities

jsmassa:
#### SUMMARY
Extends the benchmark cli to adjust 
1) the configurations for which the benchmarks are run and
2) the statistics shown in the comparison table

Also adds some logging information and removes the laziness of benchmark executions.


kordano (approved):
Looks good!

PR #544, opened 2022-06-08
Require datahike.datom in schema namespace

yflim:
#### SUMMARY
So that `schema` is self-contained and can be required on its own. Without it, a ClassNotFoundException is thrown when requiring `schema` without first requiring `datom`.

##### Bugfix
- [ ] Integration tests added
- [x] Formatting checked
- [ ] `CHANGELOG.md` updated


kordano (approved):
Looks good, thanks @yflim!

PR #546, opened 2022-06-24
add first draft for the clerks notebooks

### LINKED ISSUE #545: Improve examples

The examples should be overhauled. Probably a Clerk-Workbook would be good. We need to find a proper dataset to showcase.

- tuples
- entity-spec
- types of values/symbol-fn
- tx-metadata
- datoms/q/pull
- history/no-history-attribute
- different configs
- different databases as input

Clerk?

Krelc

kordano:
Drafts out the alternative solution for the examples folder by using [clerk](https://github.com/nextjournal/clerk) instead of plain namespaces and local comments. I moved the whole `/examples` project into simple namespaces in `/notebooks`.  Additionally I converted the `time_travel.md` from `/docs` into a clojure document to showcase how we could also move this kind of documentation to clerk.
Fixes #545.


TimoKramer:
closing this for now... we should work on this soon

PR #550, opened 2022-07-05
Remove unused query-v3 namespaces

jsmassa:
The namespace datahike.query-v3 looks like it is at an experimental stage and has never been used. 
It's time to finally delete it and its corresponding test namespace.


kordano (approved):
We didn't use these namespaces anyway, right?

Issue #525, opened 2022-04-27
DB namespace refactoring / splitting


jsmassa:
### Describe the feature you would like to request

The db namespace has grown to have more than 2000 lines, which makes analytics tools slow. It also contains code concerned with different topics, so it would make sense to split it up into multiple namespaces.

Also the `transact-tx-data` function is very large and contains different concepts of entities (map vs. vector). This could easily be split as well.

### Describe the solution you would like

- Split the namespace, e.g. in `transaction` `search`, `db-record`, `protocols`, ...
- Break up the `transact-tx-data` function into smaller pieces
- Think about using multimethods for attribute-references etc.

### Describe alternatives you've considered

Live with it as is.

PR #555, opened 2022-08-04
db.clj refactoring

jsmassa:


jsmassa:
Closes #525

kordano (approved):
Looks good from top level, let's roll it out. I fear that all the other PRs will have a bad time rebaseing to this version but better now than later.

kordano:
Could you check if that works with the latest rebase as well?

whilo:
This looks good to me as well, but it is a bit difficult to see whether the functionality has changed somewhere or whether functions have just been pulled out to different namespaces. I think it is the latter and in this case everything should be fine. 

It is a bit annoying that I have to rebase my two open PRs for the GraalVM support and the global address space again after this, but hopefully it won't cost me more than a few hours. Ideally we should settle PRs affecting functionality before doing refactorings like this and plan this a bit ahead to avoid such rebases.

PR #556, opened 2022-09-06
Add config tests to workflow

jsmassa:
Most of the tests are written only for the default configuration. With PR #503 some of the default parameters will be dynamic and can be controlled with the test profile. This PR adds separate workflows for testing the different indexes.

Blocked by #503


jsmassa:
Solved by #580

PR #559, opened 2022-09-19
Fix current value filter

### LINKED ISSUE #558: [Bug]: As-of-Databases show wrong values when old datom gets retracted

0.5.1511

11.0.13 2021-10-19

Ubuntu

```
{:store              {:backend :mem}
                                            :keep-history?      true
                                            :schema-flexibility :write
                                            :attribute-refs?    false}
```

### Describe the bug

For `cardinality/one` attributes, only the last datom gets considered when looking for a current value in as-of-databases. This leads to values being overlooked sometimes when an old value gets retracted after a new value has been added.

The last non-retracted value gets found.

```
(let [schema [{:db/ident       :name
                   :db/unique      :db.unique/identity
                   :db/valueType   :db.type/string}
                  {:db/ident       :aka
                  :db/valueType   :db.type/string}]
          conn (setup-db cfg)
          _ (d/transact conn schema)
          {:keys [tx-data] :as tx-report} (d/transact conn [{:name "Michal" :aka "Tupen"}])
          _ (println tx-report)
          michal (:e (first (filter #(= "Michal" (:v %)) tx-data)))
          {{:keys [db/current-tx]} :tempids} (d/transact conn [[:db/retract michal :aka "Tupen"]
                                                               [:db/add michal :aka "Devil"]
                                                               [:db/retract michal :aka "Tupen"]])
          _                                  (d/transact conn [[:db/retract michal :aka "Devil"]])
          as-of-db (d/as-of @conn current-tx)]
(d/pull as-of-db [:aka] michal))) ;; Should be {:aka "Devil"} but is nil
```

jsmassa:
Fixes #558

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -339,18 +339,26 @@
   #?(:cljs (instance? js/Date d)
      :clj (instance? Date d)))
 
-(defn get-current-values [rschema datoms]
-  (->> datoms
+(defn get-current-values [rschema history-datoms]
+  (->> history-datoms
        (group-by (fn [^Datom datom] [(.-e datom) (.-a datom)]))
        (mapcat
-        (fn [[[_ a] entities]]
+        (fn [[[_ a] datoms]]
           (if (contains? (get-in rschema [:db.cardinality/many]) a)
-            (->> entities
-                 (group-by #(.v %))
-                 vals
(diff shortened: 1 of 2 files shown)


kordano (approved):
Very nice, thanks!

PR #561, opened 2022-10-04
Fix as-of-filter for attribute references

### LINKED ISSUE #560: [Bug]: Property cardinality/many' gets ignored in as-of-databases

0.5.1512

openjdk 11.0.13 2021-10-19

Ubuntu

```
{:store              {:backend :mem}
 :keep-history?      true
 :schema-flexibility :write 
 :attribute-refs?    true}
```

### Describe the bug

Querying a as-of-database with attribute references returns only a single option for attributes with `:db.cardinality/many`

Should return all options.

```
(let [schema [{:db/ident       :name
                     :db/unique      :db.unique/identity
                     :db/valueType   :db.type/string}
                      {:db/ident       :aka
                       :db/cardinality :db.cardinality/many
                       :db/valueType   :db.type/string}]
              conn (setup-db (assoc keyword-cfg :attribute-refs? true))
              _ (d/transact conn schema)
              {tx-data                 :tx-data
               {:keys [db/current-tx]} :tempids} (d/transact conn [{:name "Michal" :aka ["Tupen" "Devil"]}])
              michal (:e (first (filter #(= "Michal" (:v %)) tx-data)))
              as-of-db (d/as-of @conn current-tx)]
           (d/pull as-of-db [:aka] michal) ;; returns ["Devil"], should be ["Tupen" "Devil"]
))
```

jsmassa:
Fixes #560

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -69,7 +69,7 @@
            :format {:extra-deps {cljfmt/cljfmt {:mvn/version "0.8.0"}}
                     :main-opts ["-m" "cljfmt.main" "check"]}
 
-           :ffix {:extra-deps {cljfmt/cljfmt {:mvn/version "0.9.0"}}
+           :ffix {:extra-deps {cljfmt/cljfmt {:mvn/version "0.8.0"}}
                   :main-opts ["-m" "cljfmt.main" "fix"]}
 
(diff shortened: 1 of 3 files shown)


kordano:
Good catch! :)

PR #564, opened 2022-10-09
Add some missing information to schema docs

JohnnyJayJay:
#### SUMMARY
This PR adds some more information to the `schema.md` documentation file. In its current state, the file can be misleading with regards to the set of features supported by datahike. I stumbled across this today when I made a schema that used tuple attributes and couldn't find a mention of this in this documentation file. The way it's described there led me to believe that they're not supported (yet), even though they are.

The only thing I've done here is add tuple-related optional key descriptions/mentions as well as `:db.isComponent` to the schema description section and `:db.type/tuple` to the table of supported types.

#### Checks
Neither a bugfix nor a feature, so no predefined checks.

#### ADDITIONAL INFORMATION
/

### DIFF
--- a/doc/schema.md
+++ b/doc/schema.md
@@ -88,6 +88,11 @@ Additionally, the following optional attributes are supported:
   the value for this attribute with upsert enabled)
 - `db/index`: indicates whether an index for the attribute's value should be
   created as a boolean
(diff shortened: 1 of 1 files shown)


jsmassa (approved):
Thank you for the addition to the docs, we really appreciate this :)

PR #568, opened 2022-10-24
336 native image cli

TimoKramer:
CI plus rewritten bash-script to show proper error when native-image fails


whilo (approved):
Looks good, thank you!

PR #596, opened 2023-01-11
Add specs for ns datahike.api

jsmassa:
Renewal of PR #330


test/datahike/test/validation_test.cljc
```diff
@@ -30,7 +30,7 @@
       [{:db/id -1 :profile #"regexp"}])
 
     (is (thrown-with-msg? Throwable #"Unknown operation" (d/db-with db [["aaa" :name "Ivan"]])))
-    (is (thrown-with-msg? Throwable #"Bad entity type at" (d/db-with db [:db/add "aaa" :name "Ivan"])))
+    #_(is (thrown-with-msg? Throwable #"Bad entity type at" (d/db-with db [:db/add "aaa" :name "Ivan"]))) ;; todo: how to test incorrect spec now?
```
jsmassa, line 33: At the moment the specs are being checked when the tests are run, so `db-with` already fails on the specs and it doesn't even get to the point where it would throw "Bad entity type". 

The question is now if we want to to keep on checking the specs for testing or we should  be able to have tests like this one here, which is good to have for documentation purposes in order to see what's happening.

jsmassa: But if we were not using the specs in the tests the question would be where and how else we would use them. One option would be in development environments. Probably not production though as is would slow down the code, right?

whilo: Yes, I think this is a common way to enforce contracts/spec like that, to activate them in development and debugging and turn them off in production. It really depends though also on the production environment, since they guard an interface to external access here. The main direct benefit of having specs at all in my opinion is to have automatically checked documentation for what the inputs should be.

whilo: I think this here is an edge case where we would have two guards. Ideally we could just tell that this spec check should be always on, not sure whether this is possible.

jsmassa: Hm, I can't find any documentation about how it would be possible to automatically turn on instrumentation for development. It looks like it is always turned on with calling `s/instrument` in the repl. There are libraries that are doing it differently, but they might have less features. 

I would suggest removing the instrumentation for the tests for now and instead use the specs for generative testing in the future.

kordano: A way could be to user `https://github.com/gfredericks/user.clj` together with the dev build and add `(t/instrument)` to the existing user file. I would anyway move to this file for experimental code from the old sandbox file.

src/datahike/api.cljc
```diff
@@ -85,12 +108,22 @@
   create-database
   dc/create-database)
 
+(s/fdef
+  delete-database
+  :args (s/alt :config spec/SConfig
+               :deprecated/uri string?
+               :nil (s/cat)) ; Picks up the default database
+  :ret (s/nilable map?)) ; TODO: should this be nilable?
```
kordano, line 116: I think so, or should we something like a boolean?

src/datahike/api.cljc
```diff
@@ -180,20 +213,37 @@
...
+  :ret #(s/valid? spec/STransactionReport @%))
 (def ^{:arglists '([conn tx-data tx-meta])
        :no-doc    true}
   transact!
   dc/transact!)
 
+(s/fdef
+  load-entities
+  :args (s/cat :conn spec/SConnectionAtom :txs spec/STransactions)
+  :ret #(s/valid? spec/STransactionReport @%)) ; This returns a throwable promise, so we have to dereference it..
```
kordano, line 228: At some point I would drop this function anyway since it will be superceded by the new migration tools.

src/datahike/api.cljc
```diff
@@ -218,6 +273,11 @@
                                                                 {:db/id 2, :name \"Oleg\"}]"}
   pull-many dp/pull-many)
 
+(s/fdef
+  q
+  :args (s/alt :argmap (s/cat :map spec/SQueryArgs)
+               :with-params (s/cat :q (s/or :vec vector? :map map?) :args (s/* any?))) ;; TODO: the doc could show more examples with varargs
```
kordano, line 279: Agreed, the doc should be updated.

kordano (commented):
Looks good, might need another discussion on the dev integration. But then it would be ready.

src/datahike/api.cljc
jsmassa: I would be for a boolean, especially for Datomic compatibility, but that's a different discussion.

At the moment though it seems to me like the return value of delete-database isn't even necessarily a map or nil but that this depends on the implementation of the store and basically could be anything. So either we are ok with that or define it more closely and check it in the integration tests. Anyway I think that should be a separate issue.

src/datahike/api.cljc
jsmassa: I would still leave the spec in there until the function has actually been dropped.

src/datahike/api.cljc
jsmassa: I will create an issue.

src/datahike/api.cljc
jsmassa: I created discussion #599 about that so we can decide there.

src/datahike/api.cljc
jsmassa: Issue #600 covers this now.

test/datahike/test/validation_test.cljc
TimoKramer: Did you look into [unstrument?](https://github.com/jeaye/orchestra/blob/master/src/clj/orchestra/spec/test.cljc#L51) Maybe we can stop the instrumentation for certain tests?

TimoKramer:
```
(st/unstrument)
(is (thrown-with-msg? Throwable #"Bad entity type at" (d/db-with db [:db/add "aaa" :name "Ivan"])))
(st/instrument))
```
Seems to work. Don't know if this works well because of the concurrency of the tests, though.

TimoKramer (dismissed):
Did not go through in detail with the specs. But so far looks good to me. Thanks for finishing this long standing PR @jsmassa! Especially remarkable are the changes in the api-ns where you are using this raise-macro which is awesome, wondering who made this ... 🤔. I should have used this when rewriting the api-ns.

src/datahike/api.cljc
```diff
@@ -658,17 +759,20 @@
                      :args [@conn]}) ; => #{[\"Alice\" 35] [\"Bob\" 30]}"}
   as-of
   (fn [db time-point]
-    {:pre [(or (int? time-point) (date? time-point))]}
+    {:pre [(or (int? time-point) (spec/date? time-point))]}
```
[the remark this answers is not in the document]
jsmassa: Yes, I'd also say we can drop this. I think this is there mostly for documentation purposes anyway and since this will be done by the specs there isn't a reason anymore to keep it.

src/datahike/spec.cljc
```diff
@@ -0,0 +1,95 @@
...
+(defn set-of [pred] (s/every pred :kind set?))
+
+(defn date? [d]
+  #?(:cljs (instance? js/Date d)
+     :clj  (instance? Date d)))
+
+(def time-point? (s/or :int int? :date date?))
+
+;; TODO: there's lots of runtime logic to conform these transactions - it would
+;; be nice to spec them accurately, and then conform them at runtime through spec
```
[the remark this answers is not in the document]
jsmassa: Yes, I will remove it, I think we agree on not using the specs this way

src/datahike/api.cljc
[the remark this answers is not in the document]
jsmassa: I have removed the statements in the api namespace now.

src/datahike/spec.cljc
```diff
@@ -0,0 +1,95 @@
...
+(def STransactions
+  (s/coll-of (s/or :seq coll? :map map? :nil nil?)))
+
+;; TODO: deduplicate this Spec with the one in datahike.config
```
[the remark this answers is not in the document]
jsmassa: Didn't feel worth it making an issue. Simply using the spec from config namespace now.

src/datahike/spec.cljc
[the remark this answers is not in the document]
jsmassa: It's been removed

test/datahike/test/validation_test.cljc
jsmassa: Hm, that's another option. I could live with that as well, but I think I would still personally prefer removing the instrumentation from the tests entirely.

src/datahike/spec.cljc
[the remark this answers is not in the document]
TimoKramer: Alright, you didn't want to move the spec over to spec-ns?

test/datahike/test/validation_test.cljc
TimoKramer: I am fine with not instrumenting the specs but then probably no one ever uses them again:) So maybe we should have some kind of interface-test? Just an idea...

jsmassa: What do you mean by 'interface-test'? 

As mentioned above, I would like to use them for generative tests which would act like a check if the specs are correct as well as enable us to potentially discover bugs that we wouldn't have discovered otherwise. Is that what you mean as well?

src/datahike/spec.cljc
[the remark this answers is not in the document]
jsmassa: I think that topic requires a bit more discussion. There are specs sprayed all over the place, not only the config namespace, which has quite a goo amount of them. The old specs also look different then the ones in the spec namespace. So there is a conversation to be had about which scheme we want to follow to express specs and where they should live. In some instances it might make sense to have them in separate namespaces, e.g. for the index structures, but in others we might want to move them to the spec ns.

test/datahike/test/validation_test.cljc
TimoKramer: Yes, I'd say that's what I meant. But I would rephrase it a bit... the specs are basically defining our contract with the world and therefore we should keep some kind of test that this contract is not broken and works reliably.

src/datahike/spec.cljc
```diff
@@ -15,27 +15,17 @@
 
 (def time-point? (s/or :int int? :date date?))
 
-;; TODO: there's lots of runtime logic to conform these transactions - it would
-;; be nice to spec them accurately, and then conform them at runtime through spec
-(def STransactions
-  (s/coll-of (s/or :seq coll? :map map? :nil nil?)))
+(def SDB dbu/db?)
```
TimoKramer, line 18: good one

src/datahike/spec.cljc
```diff
@@ -15,27 +15,17 @@
...
-   {:name ::config
-    :keys-default ds/opt
-    :spec {:store map?
-           :name string?
-           :keep-history? boolean?
-           :schema-flexibility (s/spec #{:write :read})
-           :initial-tx STransactions}}))
+(def SConfig (s/alt :config :datahike/config
+                    :deprecated/map :deprecated/config
+                    :deprecated/uri string?))
```
TimoKramer, line 22: thanks

src/datahike/spec.cljc
```diff
@@ -93,3 +86,19 @@
...
+(def SMetrics
+  (ds/spec
+   {:name ::metrics
+    :keys-default ds/req
+    :spec {:count int?
+           :avet-count int?
+           :per-attr-counts map?
+           :per-entity-counts map?
+           (ds/opt :temporal-count) int?
+           (ds/opt :temporal-avet-count) int?}}))
```
TimoKramer, line 104: 👍

TimoKramer (dismissed):

test/datahike/test/validation_test.cljc
jsmassa: I have created issue #605 for this topic now. Feel free to change the wording or add stuff that you think is missing.

TimoKramer (approved):

PR #597, opened 2023-01-16
Fix sys attrs in history

kordano:
#### SUMMARY
Fixes system attributes not showing up with history queries.

### DIFF
--- a/test/datahike/test/api_test.cljc
+++ b/test/datahike/test/api_test.cljc
@@ -748,8 +748,8 @@
                               :db/cardinality :db.cardinality/one
                               :db/valueType   :db.type/long}]
         conn                (utils/setup-db cfg)
-        schema-on-write?   #p (= (:schema-flexibility (.-config @conn)) :write)
-        attribute-refs? #p (:attribute-refs? (:config @conn))
+        schema-on-write? (= (:schema-flexibility (.-config @conn)) :write)
+        attribute-refs? (:attribute-refs? (:config @conn))
         update-for-schema-on-write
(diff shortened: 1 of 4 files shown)


test/datahike/test/api_test.cljc
```diff
@@ -764,16 +765,28 @@
...
+                                                               (if attribute-refs?
+                                                                 79
+                                                                 0)))
+                                  :temporal-avet-count (if schema-on-write?
+                                                         (if attribute-refs?
+                                                           48
+                                                           9)
+                                                         (if attribute-refs?
+                                                           48
+                                                           0))})))
```
jsmassa, line 789: Can you name the magical numbers in this function, so we don't have to start rethinking what was meant by them? Alternatively, you could write how they come about in comments.

test/datahike/test/gc_test.cljc
```diff
@@ -43,12 +43,12 @@
         ]
     (testing "Test initial store counts."
       (is (= 1 (count (-mark (:eavt @conn)))))
-      (is (= fresh-count (count-store @conn)))
```
whilo, line 46: You should set `num-roots` to 3 here instead of messing with the counts directly since you now structurally share the initial indices and hence have only 3 of them with or without history.

whilo (dismissed):

test/datahike/test/api_test.cljc
jsmassa: Thanks for fixing this.

jsmassa (approved):

Issue #532, opened 2022-05-06
[Bug]: schema updates and removals are not consistent


kordano:
### What version of Datahike are you using?

0.5.1501

### What version of Java are you using?

openjdk version "11.0.15" 2022-04-19

### What operating system are you using?

Ubuntu 20.04.4 LTS

### What database EDN configuration are you using?

``` clojure
{:store {:backend :mem
             :id "schema_test"}
 :keep-history?      true
 :schema-flexibility :write
 :attribute-refs?    true}
```

### Describe the bug

There are inconsistencies when updating or removing the schema.

### What is the expected behaviour?

Datahike should have a clear strategy and documentation on schema updates.

### How can the behaviour be reproduced?

``` clojure
(require '[datahike.api :as d])
(def schema [{:db/cardinality :db.cardinality/one
              :db/ident       :name
              :db/index       true
              :db/unique      :db.unique/identity
              :db/valueType   :db.type/string}
             {:db/ident       :parents
              :db/cardinality :db.cardinality/many
              :db/valueType   :db.type/ref}
             {:db/ident       :age
              :db/cardinality :db.cardinality/one
              :db/valueType   :db.type/long}])

(def cfg {:store              {:backend :mem
                               :id "schema_upgrade"}
          :keep-history?      true
          :schema-flexibility :write
          :attribute-refs?    true})

(do
  (d/delete-database cfg)
  (d/create-database cfg))

(def conn (d/connect cfg))

(d/transact conn {:tx-data schema})

(d/schema @conn)

(d/datoms @conn :eavt)

(d/transact conn [{:name "Alice"
                   :age  25}
                  {:name "Bob"
                   :age  35}])


(d/transact conn [{:name    "Charlie"
                   :age     5
                   :parents [[:name "Alice"] [:name "Bob"]]}])

(d/q '[:find ?e ?n
       :where
       [?e :name ?n]] @conn)
;;=> #{[44 "Charlie"] [43 "Bob"] [42 "Alice"]}

(d/transact conn {:tx-data [[:db/retractEntity [:db/ident :name]]]})
;; => this shouldn't be possible when there is still data with that attribute!

(d/q '[:find ?e ?n
       :where
       [?e :name ?n]] @conn)
;;=> #{[17 :db.type/boolean]
  ;; [8 11]
  ;; [8 :db.install/attribute]
  ;; [40 :parents]
  ;; [20 :db.type/float]
  ;; [1 33]
  ;; [9 :db/txInstant]
[...]

PR #598, opened 2023-01-23
feat: schema migration

### LINKED ISSUE #13: Schema Migration
I saw in #4 that schema migration is not currently supported. Are there plans to support this in the future? What's the current workaround for schema migration? Load the database with the old schema, dump it to a file, delete the database, create a database with the new schema and import the data dump?

TimoKramer:
#### SUMMARY
- schema migration as norm namespace in datahike. 
- github discussion: https://github.com/replikativ/datahike/discussions/582
- fixes #13
- relates to #532 

##### Feature
- [x] Documentation added
- [x] Formatting checked


src/datahike/norm/norm.cljc
```diff
@@ -0,0 +1,83 @@
...
+
+(defn norm-installed? [db norm]
+  (->> {:query '[:find (count ?t)
+                 :in $ ?tn
+                 :where
+                 [_ :tx/norm ?tn ?t]]
+        :args  [db norm]}
+       d/q
+       first
+       some?))
```
fpischedda, line 26: If I am not mistaken, using the dot after count should return a scalar value, if so then you can get rid of the call to `first`

Example:
```
  (d/q '[:find (count ?e)
         :where [?e :weird/stuff]]
       @db)  ;; => []

  (d/q '[:find (count ?e) .
         :where [?e :weird/stuff]]
       @db)  ;; => nil
```

```suggestion
  (->> {:query '[:find (count ?t) .
                 :in $ ?tn
                 :where
                 [_ :tx/norm ?tn ?t]]
        :args  [db norm]}
       d/q
       some?))
```

TimoKramer: good catch. unfortunately I force-pushed so cannot accept your commit suggestion anymore. But thanks.

TimoKramer:
@fpischedda I am thinking since you wished for a way to avoid conflicting schema migrations with colleagues that we could have something like a tool that tracks the migrations in a separate file. Then you would get merge conflicts in case you didn't rebase the last commit to this tracking file. @kordano came up with the idea, there is a tool called atlas for sql-dbs that does this already.

fpischedda:
Hi @TimoKramer 

This is a neat idea! I will also have a look at atlas, it could be useful at $work, thanks @kordano for mentioning it 🙂

doc/schema-migration.md
```diff
@@ -0,0 +1,52 @@
...
+create a new file with an edn-extension like `001-my-first-norm.edn`. Preferably you name the
+file beginning with a number. Please be aware that the name of your file will be the id of
+your norm. Taking into account that you might create some more norms in the future
+you should left-pad the names with zeros to keep a proper sorting. Keep in mind that your
+norms are transacted sorted after your chosen ids one after another. Spaces will be replaced
+with dashes to compose the id.
+
+2. Write the transactions itself into your newly created file. The content of the file needs
+to be an edn-map with one or both of the keys `:tx-data` and `tx-fn`. `:tx-data` is just
+transaction data in the form of a vector, `:tx-fn` is a funtion that you can run during the
```
jsmassa, line 24: Typo in `funtion`

jsmassa:
I think there is something missing in the documentation that describes in which cases it makes sense to use norms for schema migration. For someone that hasn't read the conversation in this issue it doesn't become quite clear why you would need to have norms installed for something that would need to be done only once, naively assuming you only have one database of a kind.

doc/schema-migration.md
```diff
@@ -0,0 +1,52 @@
+# Schema Migration or Norms
+
+Schema migration with Datahike is the evolution of your current schema into a future schema.
+We are calling the schema migrations 'norms' to avoid misunderstandings with a migration
+from an older version of Datahike to a newer version.
+
+When we are speaking of changes to your schema, these should always add new definitions and
```
whilo, line 7: So can I just transact new schema entries at any point or do I need to worry about norms now? This is not clear from this description. I hope it is the first, since monotonic schema additions should stay painless, I think.

doc/schema-migration.md
```diff
@@ -0,0 +1,52 @@
...
+you should left-pad the names with zeros to keep a proper sorting. Keep in mind that your
+norms are transacted sorted after your chosen ids one after another. Spaces will be replaced
+with dashes to compose the id.
+
+2. Write the transactions itself into your newly created file. The content of the file needs
+to be an edn-map with one or both of the keys `:tx-data` and `tx-fn`. `:tx-data` is just
+transaction data in the form of a vector, `:tx-fn` is a function that you can run during the
+execution to migrate data from one attribute to another for example. This function needs to
+be qualified and callable from the classpath. It will be evaluated during the migration and
+needs to return transactions. These transactions will be transacted with `:tx-data` together
```
whilo, line 27: Does this mean the transaction data needs to fully fit into memory? This can easily become a problem if the database is getting big. We can probably ignore this for now, but should at least make this constraint explicit.

doc/schema-migration.md
```diff
@@ -0,0 +1,52 @@
+# Schema Migration or Norms
```
whilo, line 1: Like @jsmassa I am not familiar with the term norm and norms have more precise meaning in mathematics. It would be good to explain what the name conveys in terms of meaning or to just say that it is named like this for consistency with Datomic. To name it like this just to use a different word than migration is not ideal for defining terminology. Why can we not just call it a `schema-migration` for instance?

test/datahike/norm/resources/naming-and-sorting-test/checksums.edn
```diff
@@ -0,0 +1,12 @@
+{:001-a1-example.edn
+ "b28aa2247b363cb556b4abba3593774b6c960d5958458bcf3ed46d86e355d563ea77c66b1f4a21cb35f232227a8d4747dd7aec519bab036c706670eb5cf4a05c",
```
whilo, line 2: I personally would prefer hasch UUIDs because they are shorter and hence more readable and conflicts between them are still less likely than asteroid impacts on earth. But it is up to you. The most important thing is that these hashes are not platform or encoding dependent (hasch can operate on ASTs of Clojure, i.e. you could read in a whole file and get a hash for syntax that ignores whitespace and orderings in maps and hash maps).

whilo (changes requested):
Overall looks good, but I will still need to take a closer look after the clarifications.

doc/schema-migration.md
TimoKramer: right, that's an issue and we would have to use batching I guess.

doc/schema-migration.md
TimoKramer: I guess the term comes from [conformity](https://github.com/avescodes/conformity) probably the most widely used migration lib for datomic.

TimoKramer: I can rename it to schema-migrations though

test/datahike/norm/resources/naming-and-sorting-test/checksums.edn
TimoKramer: I used edn-hasch -> platform/hasch->str so it should be using the clojure ast, right?

doc/schema-migration.md
jsmassa: The term "conformity" describes much better what the norms are there for. Otherwise norms can be anything depending on the context.

"Schema migration" doesn't cover everything that can be achieved with the new functionality; it's just one use case.

So maybe there should be two documents: 
1) A document explaining the new functionality with the example of schema migrations
   - can be most of the current document
   - maybe called something like "Conformity using norms"
2) A document about schema migrations which would 
  - talk about the difficulties of changing the schema of a database containing data
  - explain in which case it makes sense to use norms for schema migration
  - refer to document 1) as a guide for using norms

test/datahike/norm/resources/naming-and-sorting-test/checksums.edn
whilo: Yes, nice!

doc/schema-migration.md
whilo: I think norms here correspond to normalizing/symmetrizing a space in maths and after reading the conformity docs I get better why it is named that way. I agree that a bit more documentation would be helpful. I think the approach corresponds to a meta-editing framework for Datahike. The idempotence guarantee is necessary so you can invoke the norms time and again during transactions, I suppose, but it is not clear to me yet how good of an idea it is to norm your database like this online (i.e. for each transaction that is coming in), for instance how much overhead this causes. In case we would record all incoming data to the transactor (before it is being acted on), in any append-log such as Kafka, ZeroMQ, hitchhiker-tree or replikativ with CDVCS, we could rewrite the database history starting from the inputs to the transactor (raw data and even for transactions that fail), which is more flexible in general (and best practice as "event sourcing"), but this requires full replay. It would be helpful to me if you can describe how this approach would relate to what we have here and maybe also put it in the docs.

Note also that the invariant library covers similar functionality, but it triggers a full Datalog query itself on the transaction data and db-before/db-after pair to validate that the DB is "normed". It does not do rewrites yet though. So the trigger mechanism for norms could be unified with its approach. I am happy to have a call on this if you are interested @TimoKramer. I am fine with keeping things simple for now, but I think it is at least worth considering the different options before shipping the code to users.

TimoKramer: I think you misunderstood what this is. I rewrote large parts of the doc. Maybe you can understand it better now?

bb/resources/native-image-tests/run-normtests
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -o errexit
+set -o pipefail
+
+TIMBRE_LEVEL=':fatal' bin/kaocha --focus :norm "$@"
```
jsmassa, line 6: I think you don't really need this file, do you? The norm tests should simply be able to be called with 
`bb test norm`

doc/schema-migration.md
```diff
@@ -0,0 +1,85 @@
+# Schema Migration
+
+Schema migration with Datahike is the evolution of your current schema into a future schema.
+
+## Why using the schema-migration tool?
```
jsmassa, line 5: unnecessary hyphen: `schema-migration` -> `schema migration`

jsmassa (commented):
I like it much better now. I think the "Why using the schema migration tool?" section prevents the confusion I felt in the first version for a new reader. 

The resources/.touch file slipped in again though and can be removed.

kordano (commented):
There is still an empty `resources/.touch`in there.

doc/schema-migration.md
whilo: Yes. This might be non-trivial to add right now, but it will put a limit on the size of databases you can migrate, which is not good, as you can run into dead ends with not being able to run the migration scripts any more once the database gets too large.

doc/schema-migration.md
whilo: I am ok to postpone this design discussion, since I think I am trying to approach this too generally and the main purpose here is a simple approach for schema migration. 

Being able to replay transactions and fix the schema would nonetheless be a more general approach similar to how XTDB handles this behind Kafka and event sourcing. Replaying a transaction log also can solve the memory issue as transactions are being replayed in sequence. Deriving new attributes from the data in the database is more restricted in comparison because you might have thrown away parts of the information needed in your original schema.

test/datahike/norm/norm_test.clj
```diff
@@ -0,0 +1,120 @@
+(ns datahike.norm.norm-test
```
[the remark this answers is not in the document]
TimoKramer: Thanks, I am not aware of any right now.

whilo (approved):

jsmassa (approved):

PR #601, opened 2023-02-03
Query stats

jsmassa:
New function `query-stats` returns details like tuple counts and time during execution steps of the query engine.

### DIFF
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -1,5 +1,6 @@
 (ns user
-  (:require [datahike.api :as d]))
+  (:require [datahike.api :as d]
+            [hashp.core]))
 
 (comment
 
@@ -18,7 +19,7 @@
   (def cfg {:store              {:backend :mem :id "sandbox"}
             :keep-history?      true
             :schema-flexibility :write
-            :middleware         {:query ['datahike.middleware.query/timed]}
+            ;:middleware         {:query ['datahike.middleware.query/timed]}
             :attribute-refs?    true})
 
   (def conn (do
@@ -44,7 +45,7 @@
                      :age     45
                      :sibling [[:name "Alice"] [:name "Bob"]]}])
 
-  (d/q '[:find ?e ?a ?v ?t
+  (d/query-stats '[:find ?e ?a ?v ?t
          :in $ ?a
          :where
          [?e :name ?v ?t]
--- a/src/datahike/api.cljc
+++ b/src/datahike/api.cljc
@@ -262,6 +262,10 @@
              Query passed as map needs vectors as values. Query can not be passed as list. The 1-arity function takes a map with the arguments :query and :args and optionally the additional keys :offset and :limit."}
   q dq/q)
 
+(def ^{:arglists '([query & args] [arg-map])
+       :doc "Executes a datalog query and returns the result as well as some execution details."}
+  query-stats dq/query-stats)
+
 (defmulti datoms {:arglists '([db arg-map] [db index & components])
                   :doc "Index lookup. Returns a sequence of datoms (lazy iterator over actual DB index) which components
                         (e, a, v) match passed arguments. Datoms are sorted in index sort order. Possible `index` values
--- a/src/datahike/query.cljc
+++ b/src/datahike/query.cljc
@@ -40,6 +40,7 @@
 ;; Records
 
 (defrecord Context [rels sources rules consts])
+(defrecord StatContext [rels sources rules consts stats])
 
 ;; attrs:
 ;;    {?e 0, ?v 1} or {?e2 "a", ?age "v"}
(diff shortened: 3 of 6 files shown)


whilo:
Nice that you started working on this! I would suggest to track wall clock time in the realization of individual relations as well in addition to counting their input binding and output set. Other things that might be relevant is memory being used by a relation (in general all kinds of resources). That is harder to track, but the JVM has profiling functionality for instance. I think it would be good to discuss what we need and what we will use the profiling data for exactly. 

Also, I have already talked to @kordano about this, I am not sure whether introducing a middleware just to wrap the query on the outside is worth it, you can use `alter-var-root`  already to instrument any function like this. The name middleware is very generic and here it is a very specific and simple wrapper for query, I would expect a middleware to bind much more tightly into a stack, as is done for instance with replikativ into kabel.

jsmassa:
I didn't implement the stats as middleware anyway because I had to go deeper into the codebase. It's triggered by a simple flag in the query map at the moment.

We can add space measurements later, but I wouldn't put a lot of faith into results of the those on the JVM.

dev/user.clj
```diff
@@ -1,5 +1,6 @@
 (ns user
-  (:require [datahike.api :as d]))
+  (:require [datahike.api :as d]
+            [hashp.core]))
```
kordano, line 3: I assume that the user needs to eval that manually.

src/datahike/query.cljc
```diff
@@ -48,11 +49,49 @@
 ;; or [ (Datom. 2 "Oleg" 1 55) ... ]
 (defrecord Relation [attrs tuples])
 
+;; Main functions
+
+(defn normalize-q-input
```
kordano, line 54: Would be good to have a unit test for this function.

src/datahike/query.cljc
```diff
@@ -937,56 +993,80 @@
...
      (let [[_ & branches] clause
-           contexts (map #(resolve-clause context %) branches)
-           rels (map #(reduce hash-join (:rels %)) contexts)]
-       (assoc (first contexts) :rels [(reduce sum-rel rels)]))
+           context' (assoc context :stats [])
+           contexts (map #(resolve-clause context' %) branches)
+           sum-rel (->> contexts
+                        (map #(reduce hash-join (:rels %)))
+                        (reduce sum-rel))]
+       (cond-> (assoc context #_(first contexts) :rels [sum-rel])
```
kordano, line 1001: You still have a `#_...` in here.

src/datahike/query_stats.cljc
```diff
@@ -0,0 +1,133 @@
...
+                       :t :t-diff :clauses]))
+  ([stats cols]
+   (->> stats
+        extend-stats
+        (hash-map :branches)
+        (tree-seq map? :branches)
+        rest
+        (map (apply juxt cols)))))
+
+(comment (->> [{:rels [{:rows 3, :bound #{'?n}}],
```
kordano, line 86: This comment could be removed.

src/datahike/api.cljc
```diff
@@ -262,6 +263,10 @@
              Query passed as map needs vectors as values. Query can not be passed as list. The 1-arity function takes a map with the arguments :query and :args and optionally the additional keys :offset and :limit."}
   q dq/q)
 
+(def ^{:arglists '([query & args] [arg-map])
```
kordano, line 266: We might need some info on the execution details since it is not so clear what the different keywords mean. Maybe an example and then some explanations in the docs would be helpful.

test/datahike/test/utils.cljc
```diff
@@ -26,3 +26,23 @@
...
+        _ (d/create-database config)
+        conn (d/connect config)]
+    (d/transact conn test-data)
+    conn))
+
+(defn teardown-db [conn]
+  (d/release conn)
+  (d/delete-database (:config @conn)))
+
+(defn with-db
```
kordano, line 41: I like these utilities, at some point we would need again some refactoring of our test setup.

dev/user.clj
jsmassa: Added the user.clj library

src/datahike/api.cljc
jsmassa: It's exactly the same as for q, so I don't think we need extra documentation. I added a note about that behaviour in the docstring though.

src/datahike/query.cljc
jsmassa: I removed it now

src/datahike/query_stats.cljc
jsmassa: Removed.

src/datahike/query.cljc
jsmassa: Started doing this, seems I am having some trouble with my local setup atm though, so I have to solve this first.

whilo:
I think an easy way to do approximate space measurements would be to serialize each context after it has been created and measure its byte size (the simplest option is `pr-str` with `count`, which might be a bit misleading because it prints numbers instead of storing them efficiently, CBOR/fressian would be better). This should happen after timing of course and probably be optional because of the slow down, but it would give a good sense for memory usage of each join. I think this will be important to know as maintaining memory bounds is necessary in many applications, while response time might not be most critical. It can be done later, but rendering more of the behaviour visible now will help us in making design decisions.

src/datahike/query.cljc
jsmassa: The tests are finished now.

kordano (approved):

PR #602, opened 2023-02-17
WIP: Rule cache

jsmassa:
Add caching of results for queries for performance enhancement


whilo:
Thanks for exploring this! Closing as superseded: the query engine this patches was rewritten wholesale (query planner, #813–#883), main now ships a built-in LRU query-result cache (`datahike.query/*query-result-cache?*` with size/weight knobs), and rule bodies are routed through the IR pipeline. The middleware namespace this branch modifies no longer exists on main.

PR #606, opened 2023-02-27
test: spec profile

jsmassa:
#### SUMMARY
Follow-up for #596

Introduces kaocha profile that can run with specs enabled


whilo (approved):
Looks good to me. Thanks for factorizing it this way, I really like that the specs can be turned off from the outside for all tests!

PR #608, opened 2023-03-03
chore: update deps 

jsmassa:
Changes:
- updates dependencies
- adds babashka task for deps update


deps.edn
```diff
@@ -67,19 +67,19 @@
...
                                     incanter/incanter-charts {:mvn/version "1.9.3"}}
                        :main-opts ["-m" "benchmark.cli"]}
 
-           :format {:extra-deps {cljfmt/cljfmt {:mvn/version "0.8.0"}}
+           :format {:extra-deps {cljfmt/cljfmt {:mvn/version "0.9.2"}}
                     :main-opts ["-m" "cljfmt.main" "check"]}
 
-           :outdated {:extra-deps {com.github.liquidz/antq {:mvn/version "2.2.983"}}
-                      :main-opts ["-m" "antq.core"]}}}
+           :outdated {:extra-deps {com.github.liquidz/antq {:mvn/version "2.2.992"}}
```
jsmassa, line 84: It would be nice to run this either in the ci pipeline or as a regular task somehow.

deps.edn
```diff
@@ -67,19 +67,19 @@
...
+                      :main-opts ["-m" "antq.core" "--exclude=com.github.liquidz/antq@2.2.999"]}}}
```
jsmassa, line 85: We could also hold the excluded packages in the `config.edn.` I only put it here for now because it seemed simpler. 

Can also get attached to the packages themselves, see [here](https://github.com/liquidz/antq/blob/main/doc/exclusions.adoc)

bb/deps.edn
```diff
@@ -1,8 +1,8 @@
 {:paths ["src" "resources"]
  :deps {org.babashka/spec.alpha {:git/url "https://github.com/babashka/spec.alpha"
-                                 :git/sha "644a7fc216e43d5da87b07471b0f87d874107d1a"}
+                                 :git/sha "951b49b8c173244e66443b8188e3ff928a0a71e7"}
         ;; https://github.com/babashka/tools.bbuild
-        io.github.babashka/tools.bbuild {:git/sha "0a4959d2b1147f79e9fb37e0243727390b8ba2b3"}
+        io.github.babashka/tools.bbuild {:git/sha "a3c0723b6250756eaeed532ade5d6474df24e80a"}
         ;; https://github.com/borkdude/gh-release-artifact
-        io.github.borkdude/gh-release-artifact {:git/sha "05f8d8659e6805d513c59447ff41dc8497878462"}
+        io.github.borkdude/gh-release-artifact {:git/sha "db5e79559fdbfa009ed4a0921079e20e1e76b269"}
```
jsmassa, line 7: I cannot decide if it would be too dangerous to upgrade the build packages too often or if it is not too much of a problem if a build or release fails from time to time.

bb.edn
```diff
@@ -22,7 +22,9 @@
          
          ffix {:doc "Format source files" 
                :task (clojure "-M:ffix")}
-
+         
+         dupgrade {:doc "Upgrade all clojure deps"
+                   :task (clojure "-M:outdated" "--upgrade" "--force")}
```
jsmassa, line 27: We could also have this without `--force`, then each package upgrade has to be confirmed

bb/deps.edn
jsmassa: Ok, I already had to exclude the tools.bbuild because the most recent version made the builds fail.

deps.edn
kordano: @TimoKramer, any insights on this?

kordano (approved):

deps.edn
TimoKramer: I don't see how we would be running this during CI. It would fail all the time and updating deps shouldn't be part of each random PR but should be it's own PR, I think. It would be much better to have kind of a bot to open PRs on this. Maybe I can build something...

TimoKramer (approved):

deps.edn
TimoKramer: https://github.com/nnichols/clojure-dependency-update-action/blob/master/dependency-check.sh

jsmassa: By now, I would also tend to rather have a separate task for this than to require an update with each PR. 

It was annoying enough to try to do it perfectly in this PR and I can imagine it would be even more annoying if there were major code changes involved that you wanted to get in asap.

jsmassa: Can we use GitHub Actions alongside Circleci?

TimoKramer: sure

jsmassa: Nice, then it shouldn't be too hard to set this up.

TimoKramer:
Closing because we should use a bot to check for updatability regularly and we won't need an alias for it. I personally prefer `neil` anyway.

PR #614, opened 2023-03-10
chore: remove cljs dep from prod

jsmassa:
#### SUMMARY
Move ClojureScript dependency to `:cljs` alias.
Adds `data.json` dependency for the native cli. It seems it had been provided via the ClojureScript library.

#### ADDITIONAL INFORMATION
- The ClojureScript dependency is quite large
  - see discussion #613
- There is currently no ClojureScript support in Datahike (version 0.6.1538)
- Most Clojure/ClojureScript expect ClojureScript library to be provided (e.g. [datascript](https://github.com/tonsky/datascript/blob/1a405721cb2226a28678066b176a5326c264153e/project.clj#L11), [environ](https://github.com/weavejester/environ/blob/aa90997b38bb8070d94dc4a00a14e656eb5fc9ae/environ/project.clj#L8))

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,12 +1,10 @@
 {:deps {org.clojure/clojure                         {:mvn/version "1.11.1"}
-        org.clojure/clojurescript                   {:mvn/version "1.11.4"}
         io.replikativ/hasch                         {:mvn/version "0.3.7"}
         io.replikativ/hitchhiker-tree               {:mvn/version "0.2.222"}
         io.replikativ/incognito                     {:mvn/version "0.3.66"}
         io.replikativ/konserve                      {:mvn/version "0.7.285"}
         persistent-sorted-set/persistent-sorted-set {:mvn/version "0.2.1"}
         environ/environ                             {:mvn/version "1.2.0"}
-        cheshire/cheshire                           {:mvn/version "5.11.0"}
         com.taoensso/timbre                         {:mvn/version "5.2.1"}
         io.replikativ/superv.async                  {:mvn/version "0.3.43"}
         io.replikativ/datalog-parser                {:mvn/version "0.2.25"}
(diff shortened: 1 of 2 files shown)


whilo (changes requested):
Cheshire is used by libdatahike, so you should probably move it as a dependency into its profile, otherwise looks good. https://github.com/replikativ/datahike/blob/main/libdatahike/src/datahike/impl/libdatahike.clj#L3

jsmassa:
We are getting into trouble building the api docs when moving the `cheshire` lib back to an alias. Unless it is an big dependency I wouldn't move it away from the common libs because of this issue. We could introduce a `projects` directory similar to the way it is in polylith to isolate the cli namespace and then we can easily drop the package that is only needed there but as long as it is mixed in with the rest of datahike it can be quite confusing not to be able to load the namespace with the common dependencies.

deps.edn
```diff
@@ -1,20 +1,21 @@
...
         io.replikativ/datalog-parser                {:mvn/version "0.2.25"}
         io.replikativ/zufall                        {:mvn/version "0.1.0"}
         junit/junit                                 {:mvn/version "4.13.2"}
         medley/medley                               {:mvn/version "1.4.0"} 
         metosin/spec-tools                          {:mvn/version "0.10.5"}
-        mvxcvi/clj-cbor                             {:mvn/version "1.1.0"}}
+        mvxcvi/clj-cbor                             {:mvn/version "1.1.0"}
+        ;; datahike cli
+        org.clojure/data.json                       {:mvn/version "0.2.6"} 
+        cheshire/cheshire                           {:mvn/version "5.11.0"}}
```
TimoKramer, line 19: these lines are not needed anymore, just rebase and they are gone into the alias

TimoKramer (approved):
thanks, merge it please. problem is though, that in each replikativ-dep there is clojurescript as dep :)

PR #615, opened 2023-03-10
chore: remove incognito dependency

jsmassa:
#### SUMMARY
Removes incognito library which is not being used anymore.


whilo (dismissed):

whilo:
This also needs to be removed from konserve, if you want to get it out of the dependencies completely. It is still used by the hitchhiker-tree, though, right?

whilo (approved):
Thanks!

TimoKramer:
@jsmassa merge?

TimoKramer:
@whilo can you review again please? I removed clojurescript from deps and did some updates. (Unfortunately the latest releases of cheshire and persistent-sorted-set fails, so did not upgrade them)

whilo:
We should port our native code from cheshire to jsonista and remove the dependency, I think. I will update persistent-sorted-set in my next PR after the server.

whilo (approved):
Looks good.

PR #618, opened 2023-03-20
feat: improve writer latency

### LINKED ISSUE #617: Improve writer latency

The writer currently writes indices and the db record to the underlying store in sequence, which is not required.

The writer should write as much in parallel as possible.

-

whilo:
Fixes #617. This pull request changes the operations of the write process to flush and sync the dirty indices in a two stage process instead of waiting on all operations during the execution:

1. Flush all index trees and collect asynchronous operations on dirty nodes without waiting. Then wait on collected operations until all index data is written. (This ensures that no pointers used by the DB record in 2. will be dangling.)
2. Write DB record in parallel into commit log and into branch value. 

This approach reduces the transaction latency in the best case to two round trips to the underlying store, which is optimal if distributed snapshot consistency needs to be preserved. Otherwise other processes could read DB records that point to tree fragments that are not yet written.


src/datahike/writer.cljc
```diff
@@ -1,75 +1,143 @@
...
-                                    {:type       :writer-error-during-invocation
-                                     :invocation invocation
-                                     :connection connection
-                                     :error      e})
-                           e)))]
-           (when (some? callback)
-             (put! callback res)))
-         (recur))
-       (log/debug "Writer thread gracefully closed")))))
+  [connection write-fn-map transaction-queue-size commit-queue-size commit-wait-time]
```
whilo, line 35: This is the main added logic. It first transacts in a loop and then commits in the second loop. Transactions are grouped in commits and they all see the same db-after containing their transaction and their individual datoms added and retracted.

src/datahike/index/persistent_set.cljc
```diff
@@ -147,13 +147,13 @@
...
 
-(defrecord CachedStorage [store config cache stats]
+(defrecord CachedStorage [store config cache stats pending-writes]
   IStorage
   (store [_ node]
     (swap! stats update :writes inc)
     (let [address (gen-address node (:crypto-hash? config))
           _ (trace "writing storage: " address " crypto: " (:crypto-hash? config))]
-      (k/assoc store address node {:sync? true})
+      (swap! pending-writes conj (k/assoc store address node {:sync? false}))
```
whilo, line 156: Write operations are tracked here and collected by commit! later.

src/datahike/index/persistent_set.cljc
```diff
@@ -177,19 +178,19 @@
...
+                  (atom [])))
 
-(def ^:const BRANCHING_FACTOR 512)
+(def ^:const DEFAULT_BRANCHING_FACTOR 512)
 
 (defmethod di/empty-index :datahike.index/persistent-set [_index-name store index-type _]
-  (psset/set-branching-factor! BRANCHING_FACTOR)
-  (let [^PersistentSortedSet pset (psset/sorted-set-by (index-type->cmp-quick index-type false))]
-    (set! (.-_storage pset) (:storage store))
+  (let [^PersistentSortedSet pset (psset/sorted-set* {:cmp (index-type->cmp-quick index-type false)
```
whilo, line 187: This is for the new persistent sorted set version.

src/datahike/writing.cljc
```diff
@@ -86,57 +109,63 @@
...
-                                  (assoc-in tx-report [:tx-meta :db/commitId]
-                                            (get-in db [:meta :datahike/commit-id])))]
-     (reset! connection db)
-     (doseq [[_ callback] (some-> (:listeners (meta connection)) (deref))]
-       (callback tx-report))
-     tx-report)))
+  (let [{:keys [hash max-tx max-eid meta]} db]
+    (uuid [hash max-tx max-eid meta])))
+
+(defn commit!
```
whilo, line 127: All durable changes happen here.

src/datahike/writing.cljc
```diff
@@ -86,57 +109,63 @@
...
+                  (<?- branch-op)
+                  db)))))
+
+(defn add-commit-meta! [connection commit-db]
+  (let [{{:keys [datahike/commit-id datahike/parents]} :meta} commit-db]
+    (swap! connection #(-> %
+                           (assoc-in [:meta :datahike/parents] parents)
+                           (assoc-in [:meta :datahike/commit-id] commit-id)))))
+
+(defn update-connection! [connection tx-data tx-meta update-fn]
```
whilo, line 154: Transactions (transact and load-entities) happen here.

TimoKramer (commented):
so far looks good to me. one thing I do not understand is, why you're using two loops in the writer...

src/datahike/index/persistent_set.cljc
```diff
@@ -166,6 +166,7 @@
     (if-let [cached (wrapped/lookup cache address)]
       cached
       (let [node (k/get store address nil {:sync? true})]
+        (assert (not (nil? node)) "Node not found in storage.")
```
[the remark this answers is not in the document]
whilo: This should never happen. All the nodes we refer to in our addresses/pointers must always exist, otherwise the store got corrupted or a write operation was inconsistent (e.g. the underlying store did not ensure its durability or mixed up the causal order of events [I think Redis, does, but it does not immediately store durably, so something like this could go wrong in a faulty store backend]). If your comment is about the assertion, yes I will change that to an Exception now.

But we also could allow it and retry. In this case you would be able to sync the store from another system out of order (e.g. through datsync, rsync or file copying or whatever) and your operations would just stall instead of break. But I think it is better to just copy everything except for the roots first and overwrite the roots last in an atomic operation (in the same order commit writes), that way you can sync without interrupting operations at all.

TimoKramer (dismissed):
I think it's good.

src/datahike/writing.cljc
```diff
@@ -86,57 +109,63 @@
...
+                      cid           (create-commit-id db)
+                      db            (-> db
+                                        (assoc-in [:meta :datahike/commit-id] cid)
+                                        (assoc-in [:meta :datahike/parents] parents))
+                      db-to-store   (db->stored db true)
+                      _             (<?- (flush-pending-writes store sync?))
+                      commit-log-op (k/assoc store cid db-to-store {:sync? sync?})
+                      branch-op     (k/assoc store (:branch config) db-to-store {:sync? sync?})]
+                  (<?- commit-log-op)
+                  (<?- branch-op)
```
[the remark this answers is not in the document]
whilo: This is waiting for the write operations in the asynchronous case and a no-op in the synchronous case (because the values are already written then). It also will re-throw exceptions that would be in either of these channels.

src/datahike/writing.cljc
```diff
@@ -86,57 +109,63 @@
...
-                (if-not (keyword? p) p
-                        (let [{{:keys [datahike/commit-id]} :meta :as old-db}
-                              (k/get store p nil {:sync? true})]
-                          (when-not old-db
-                            (dt/raise "Parent does not exist in store."
-                                      {:type   :parent-does-not-exist-in-store
-                                       :parent p}))
-                          commit-id))))))
+                (do
+                  (assert (not (nil? p)) "Parent cannot be nil.")
```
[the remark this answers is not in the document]
whilo: thanks, fixed.

TimoKramer (approved):

PR #620, opened 2023-03-23
feat: clojure backwards compatibility

### LINKED ISSUE #619: Clojure backward compatibility

Compatibility of datahike with earlier versions of clojure (and clojurescript).

Adding tests using earlier versions of clojure, for example by adding a test config for kaocha.

Only have datahike compatible with the most recent version as it is right now. 

This restricts the use of datahike though to projects where using the latest version can be easily achieved. 

For example, we had problems reported after PR #601 had been merged into main because `abs` had just recently been included into the core project.

jsmassa:
#### SUMMARY
Fixes #619.

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [x] Formatting checked

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,8 +138,34 @@ jobs:
       - attach_workspace:
           at: /home/circleci
       - run:
-          name: Run unittests for hitchhiker-tree index
+          name: Run unittests with specs activated
           command: bb test specs
+          no_output_timeout: 5m
+      - save_cache:
+          key: deps-{{ checksum "deps.edn" }}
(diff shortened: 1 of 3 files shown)


whilo:
Thanks for this work! Closing as superseded: the `:1.9`/`:1.10` override aliases landed on main independently (deps.edn), but main now pins Clojure 1.12 and three years of engine work have never been tested against 1.9/1.10, so the compat claim here no longer holds. If 1.10 CI coverage is still wanted, it's a small fresh PR adding CI jobs on top of the aliases that already exist.

PR #631, opened 2023-07-17
move norm tx out of log

kordano:
#### SUMMARY
- fixes norms not applying when log level is set above info


TimoKramer (approved):
I really don't get how the threading macro into the log is avoiding the transactions to be applied (because of lazyness)?? Couldn't reproduce now but believing you and the change is fine, who needs the tx-report in the info-log anyway...

PR #637, opened 2023-08-23
I/624/prep lib is broken

### LINKED ISSUE #624: [Bug]: prep-lib is broken

main

19

Linux

-

### Describe the bug

When using datahike as a lib I need to run `clj -X:deps prep` to compile the java classes. This throws `Prep function could not be resolved: tools.build/jcompile` because the `clj -T:build jcompile` functionality seems to be broken. Tried from datahike-server repository.

In the datahike-folder when cloned to disk `clj -X:deps prep` does nothing but `clj -T:build compile` throws `java.lang.Exception: Cyclic load dependency: [ /clojure/main ]->/clojure/spec/alpha->[ /clojure/main ]->/clojure/core/server`

`clj -X:deps prep` should compile the java code [like it is described here.](https://clojure.org/reference/deps_and_cli#prep)

`clj -X:deps prep`

oelrich:
#### SUMMARY

This adds a trigger for the babashka build system and now `clj -X:deps prep` should work for at library using datahike.

##### Bugfix
Fixes #624

### DIFF
--- a/build.clj
+++ b/build.clj
@@ -1,0 +1,5 @@
+(ns build
+  (:require [clojure.java.shell :refer [sh]]))
+
+  (defn bb-trigger [_input]
+    (sh "bb" "jcompile"))
--- a/deps.edn
+++ b/deps.edn
@@ -18,8 +18,8 @@
  :paths ["src" "target/classes"]
 
  :deps/prep-lib {:ensure "target/classes"
-                 :alias :build
-                 :fn jcompile}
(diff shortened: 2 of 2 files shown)


TimoKramer:
@jsmassa can you review this? I am pretty sure we were not using the build-alias for anything else so it should be good, right?

TimoKramer (approved):

TimoKramer:
Thanks @oelrich

Issue #642, opened 2023-09-19
[Bug]: NullPointerException trying to transact from CLI


acobster:
### What version of Datahike are you using?

0.6.1547

### What version of Java are you using?

openjdk 11.0.20.1

### What operating system are you using?

Ubuntu 20.04.1

### What database EDN configuration are you using?

```
{:store  {:backend :file
          :path "/home/tamayo/Downloads/testdb"
          :config {:in-place? true}}
 :keep-history? true
 :schema-flexibility :read}
```

### Describe the bug

After downloading and extracting the CLI, I configured and created a database and then attempted to transact to it. Instead it throws a `NullPointerException` and prints a stack trace.

### What is the expected behaviour?

Following the [example usage](https://cljdoc.org/d/io.replikativ/datahike/0.6.1547/doc/command-line-interface#example-usage), I expected the `./dhi transact ...` command to succeed. :)

### How can the behaviour be reproduced?

Download and extract datahike-0.6.1547-linux-x86_64.zip, create a database, and transact to it:

```
⚡  ~/Downloads $ ./dhi create-database testdb.edn                                                                                                                                                                 
{:keep-history? true, :search-cache-size 10000, :index :datahike.index/persistent-set, :store {:path "/home/tamayo/Downloads/testdb", :scope "127.0.1.1", :backend :file, :config {:in-place? true}}, :store-cache-
size 1000, :attribute-refs? false, :writer {:backend :self}, :crypto-hash? false, :schema-flexibility :read, :branch :db}                                                                                          
⚡  ~/Downloads $ ./dhi database-exists testdb.edn                                                                                                                                                                 
true                                                                                                                                                                                                               
⚡  ~/Downloads $ ./dhi transact db:testdb.edn '[[:db/add -1 :name "Coby"]]'                                                                                                                                       
Exception in thread "main" java.lang.NullPointerException                                                                                                                                                          
        at clojure.core$deref_future.invokeStatic(core.clj:2317)                                                                                                                                                   
        at clojure.core$deref.invokeStatic(core.clj:2337)                                                                                                                                                          
        at datahike.writer$transact_BANG_.invokeStatic(writer.cljc:105)                                                                                                                                            
        at datahike.api$transact.invokeStatic(api.cljc:241)                                                                                                                                                        
        at datahike.cli$_main.invokeStatic(cli.clj:192)                                                                                                                                                            
        at datahike.cli$_main.doInvoke(cli.clj:146)                                                                                                                                                                
[...]

PR #643, opened 2023-09-21
Misc. docs improvements

acobster:
#### SUMMARY

Fixes to various rough edges in the docs.

- Addresses incorrectly documented `transact` CLI command #642
- Consolidates documentation about a Datahike Babashka pod - the stub `babashka_pod.md` is currently shadowing the other one (see left nav on the [docs website](https://cljdoc.org/d/io.replikativ/datahike/0.6.1545/doc/datahike-babashka-pod)), hiding lots of good info!
- Fixes a dead link to the "Roadmap"

#### Checks

Docs only, no code changes.


doc/cli.md
```diff
@@ -37,11 +37,14 @@ $ datahike query '[:find ?n . :where [?e :name ?n]]' db:myconfig.edn
 "Linus" # prints the name
 ```
 
+Note that the `conn:<file>` argument to `transact` comes before the transaction
+value(s), whereas the `db:<file>` argument to `query` comes after the query value,
```
whilo, line 41: The reason for this is that you can pass more than one db, in fact an arbitrary number of different data sources into the query engine. This is a core benefit of using datahike, so maybe we should also highlight that here.

whilo (approved):
Thank you! Overall looks good, feel free to address the minor comment I added if you have time.

doc/cli.md
acobster: Good point, I emphasized the connection by combining the two paragraphs. Let me know what you think. Also fixed a couple punctuation errors that I noticed.

whilo: You can pass also files with `edn:/some/path/file.edn` or `json:/some/path/file.json`. My idea in general is to keep the way inputs are represented medium specific, while the API should always be the same. For a command line client it was necessary to provide a small DSL for file loading and dereferencing. 

This is also mentioned here:
https://github.com/replikativ/datahike/blob/main/src/datahike/cli.clj#L20

whilo: As a sidenote: Files are fairly reasonable as an abstraction on Unix, but maybe there are better ways to design this interface.

acobster: Hmmm interesting. Well now that you got me thinking about it, one thing this DSL rules out is using Bash [process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html#Process-Substitution) to specify file contents ad-hoc, like:

`datahike transact <(echo '{:store {:backend :mem :id "mydb"}}') ...`

I don't know when I would use that over a regular file, and if you wanted to support multiple formats with it you would need to pair it with an option or something (`datahike transact --edn <(...)`). Just thinking out loud, since you mention Unix and files.

whilo: Transact does support to read in transaction data from STDIN already, but that is only for transact. I need to think about it more. I think you could pipe through files, right?

whilo (approved):
Thanks so much for these fixes!

whilo:
@acobster I decided to merge this for now and we should discuss further improvements in follow up issues or on slack if you are down.

PR #644, opened 2023-09-24
fix: commit classes for git deps

groundedsage:


whilo:
Nice, thanks! Which Java version did you compile the class files with?

TimoKramer:
Why would you want that class files in the repo? Is there a Problem somewhere that justifies that?

whilo:
Yes, the problem is that you cannot directly use deps with git anymore, because you first need to compile the dependency locally and then point to the jar for each commit and branch. This means any non-released datahike version cannot be easily used and tested by downstream projects. Electric also commits class files like this because they barely change and are small (our Java binding is very stable).

TimoKramer:
I can not [find a class file in the electric repo](https://github.com/search?q=repo%3Ahyperfiddle%2Felectric+path%3A*.class&type=code), where can I find it? [Why is this not sufficient?](https://clojure.org/guides/deps_and_cli#prep_libs)
Also, as you pointed out, with which version are the class files compiled?

whilo:
@groundedsage Did you try to use prep-lib?

TimoKramer:
closing for now

groundedsage:
@TimoKramer classes are in this [directory](https://github.com/hyperfiddle/electric/tree/master/src/hyperfiddle/electric) and mentioned in [build.md](https://github.com/hyperfiddle/electric/blob/cc55772f18bc46373f131e092fc20055c8062b59/src-build/build.md?plain=1#L45)

I just tried `pre-lib` and this did not work. On the surface this does appear to be the correct way to handle this though. I will submit an issue for this specifically.

PR #645, opened 2023-10-10
proposal exception middleware

TimoKramer:
I don't have time to do this properly today but this is a solution to the problem that sometimes exceptions are not caught and huge responses are returned.


TimoKramer:
@whilo I think this should solve the last problems I saw

whilo (approved):
Nice work, thank you!

Issue #654, opened 2023-11-05
Ability to disable `ensure-stored-config-consistency`


alekcz:
### Describe the feature you would like to request

The `ensure-stored-config-consistency` function that validates connection creates issues when a valid connection may different names.

e.g. on JDBC these two stores are equivalent but not be deemed so. They appear different as because in one case the DB and datahike are on the same network and the in the other they are not. 
```
(def pg-cfg  {:dbtype  "postgresql"
              :jdbcUrl "postgresql://user:password@localhost/konserve"})
```

```
(def pg-cfg  {:dbtype  "postgresql"
                    :jdbcUrl "postgresql://user:password@myremoteaddress.com/konserve"})
```
A tangentially related question: with the  `ensure-stored-config-consistency` how would one go about updating the cache size?

### Describe the solution you would like

- Either the ability to disable the check.
- Or a key that can be passed in the config that can be used to validate that the user did indeed intend to connect to the db with a different looking config. 

### Describe alternatives you've considered

None really. I'm not familiar enough with Datahike to work on the internal code.

alekcz:
@whilo @TimoKramer thoughts?

whilo:
I think disabling might not be the best idea. The problem is once you connect to a store with a potentially incompatible configuration you might corrupt the database. I agree that the `jdbcUrl` should be allowed to be different for the hostname and port. Maybe we could provide a multimethod that is implemented by every store and only returns the part of the config that should be matched.

whilo:
`store-identity` is supposed to do that, I need to check again how it works for JDBC. Alternatively we could still use the random names and add them to every store and just rely on that.

alekcz:
Lemme work on making store-identity more robust.

alekcz:
@whilo I've almost completed the `store-identity` work on `datahike-jdbc`. The missing part in on the datahike side. The comparison doesn't use the `store-identity`: https://github.com/replikativ/datahike/blob/166cdbbb7031fa03e619091fb7a99c20cd550ec3/src/datahike/connector.cljc#L126C18-L126C18

whilo:
Right, I think it should break out the store config there and run store-identity on it and then equality check that.

alekcz:
Resolved by #655

PR #655, opened 2023-11-08
bugfix: use store identity in config check

alekcz:
#### SUMMARY
The `store-identity` is implemented as part of the backend protocol but is not used in assessing if configurations are equivalent. As a result two stores are functionally equivalent are not be deemed so unless they are literally equivalent. This fix uses the `store-identity` to check equivalence for configurations and connections.

e.g. on JDBC these two stores are equivalent but not be deemed so. They appear different as because in one case the DB and datahike are on the same network and the in the other they are not.
```
(def pg-cfg  {:dbtype  "postgresql" :jdbcUrl "postgresql://user:password@localhost/konserve"})
```
```
(def pg-cfg  {:dbtype  "postgresql" :jdbcUrl "postgresql://user:password@myremoteaddress.com/konserve"})
```
##### Bugfix
- [x] Fixes issue #654
- [x] Integration tests added
- [x] Formatting checked

### DIFF
--- a/src/datahike/connector.cljc
+++ b/src/datahike/connector.cljc
@@ -122,7 +122,10 @@
         [config stored-config] (if-not (= dc/self-writer config)
                                  [(dissoc config :writer)
                                   (dissoc stored-config :writer)]
-                                 [config stored-config])]
+                                 [config stored-config])
+        ;; replace store config with its identity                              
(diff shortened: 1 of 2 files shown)


alekcz:
@whilo this should address #654

whilo (approved):
Nice work. Thanks!

Issue #206, opened 2020-07-16
ClojureScript Support


samdesota:
Hey Guys,

I'm experimenting with using Datahke as a basis to build an extensible database system. As part of this, support for nodejs is important, so I've started to iterate on cljs support for datahike.

https://github.com/replikativ/datahike/compare/development...nanodevstudio:development

Wanted to see if there's any big barriers to this that I might run into or any thoughts you have around this, since I believe this is listed as goal in the README. My current approach is just repeatedly running the cljs build and tests until they all pass, squashing cljs incompatibilties as I go.

whilo:
Hey Samuel, 

that is super cool, I think we should have a chat together with @groundedSAGE . 

We definitely also want to have an extensible database system spanning devices from frontend to large scale backends, this ambition separates us from the other work happening in this design space at the moment. If I read your PR correctly you have mostly tried to port Datahike namespaces to compile again. This is a good start. We also need to do some work on the JS storage backends and on the hitchhiker-tree. After that we need to port the query engine to core.async to support asynchronous IO in ClojureScript, this is probably the biggest piece, but definitely doable.

There are a lot whitespace changes in the PR, we are discussing which standard format to use, could you keep these changes separate from the PR for now?

samdesota:
Yesterday I succeeded in getting the datahike namespaces to at least compile, your comment gives me a good roadmap, thanks.

Also disabled my autoformatter and fixed the whitespace issues, should be clearer now.

Would love to chat with you guys about datahike. One goal of mine is to have a database system which can be expanded with custom indexes for search, geo etc. while keeping everything within a core datalog-like query langage, and allowing the developer to make choices about the consistency of the various indexes for performance. 

I think CLJS support could also help with building out a really great admin UI for datahike. Looks like there's some discussion in #205

kordano:
Yeah, I'm trying to structure the requirements for the admin UI. Let's have a chat once I've done that.

whilo:
This is available as a preview in a feature branch now: https://github.com/replikativ/datahike/tree/206-cljs-support. Feedback is appreciated! Please join our discord here: https://discord.gg/kEBzMvb.

grav:
Just wanted to chime in and say that I've gotten the https://github.com/replikativ/datahike/tree/206-cljs-support branch working with NodeJS.

I used this shim: https://www.npmjs.com/package/indexeddbshim
Using Shadow-cljs, I can use it like this:

```
(ns foo
[...]

PR #662, opened 2024-02-09
In-memory cljs support.

whilo:
Addresses #206. First phase is going to be to make Datahike compile with cljs support in-memory. Compared to https://github.com/replikativ/datahike/tree/206-cljs-support this does not require a complete move to core.async yet, but will allow synchronizing indices between clojure and clojurescript peers, e.g. with replikativ. In the second phase we can then port Datahike to support both asynchronous and synchronous execution.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -32,7 +32,9 @@
 
            :1.10 {:override-deps {org.clojure/clojure {:mvn/version "1.10.0"}}}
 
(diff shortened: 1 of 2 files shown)


whilo:
For reference, there is also a partial effort here https://github.com/replikativ/datahike/pull/357

whilo:
This is now covered in a better way with #748.

PR #674, opened 2024-05-09
Refactor query and search for better performance

jonasseglare:
This code mainly refactors the namespaces `datahike.query` and `datahike.db.search` to improve performance of the query engine. This means that we remove the previous "relprod strategies" and the algorithm in this PR best resembles the "select-all" strategy that we had before. I believe there are many things that can contribute to the speed improvement, but this refactoring means that work is moved out from the innermost loop of the query engine and this loop now consists of a transduction inside `search-batch-fn`. Using transducers probably avoid the creation of lots of short-lived intermediate values and lazy sequences and seem to contribute to the performance. In some cases I use macros to move work out of loops and that comes at a cost in terms of readability. I explored simpler ways of writing the code but in the places where I do use macros to generate code to avoid work at runtime, they seem to be justified, see for example the comments in the function `single-substitution-xform`.

Here are the results. The results for this PR is labeled `Datahike new query/search` in the table below:

```
TARGET                      ABS TIME (s)   REL TIME
Some other db                      4.471         6%
Datahike 0.6.1558                 74.294       100%
Datahike 0.6.1559                 51.827        70%
Datahike new query/search         26.377        36%
```

The code to run the benchmark is found at <https://gitlab.com/arbetsformedlingen/taxonomy-dev/backend/experimental/datahike-benchmark/>.

I believe @whilo will want to review this code, he knows what it is about.


src/datahike/query.cljc
```diff
@@ -1033,191 +1055,537 @@ in those cases.
...
+                                    ~(mapv
+                                      (fn [sym] `(nth ~obj ~sym))
+                                      index-symbols)))]))
+                   (range (inc max-length)))))))
+
+(def make-basic-index-selector (basic-index-selector 5))
+
+(defn single-substitution-xform
+  "Returns a transducer that substitutes the symbols for a single relation."
+  [search-context relation-index subst-map filt-map]
```
jonasseglare, line 1311: I should think about better names instead of `subst-map` and `filt-map`. This code is a bit complicated so better names would help for the sake of readability.

whilo: Right, that helps in general, but it is also ok if they are explained where they are introduced, or renamed to shorthands in internal scopes.

src/datahike/query.cljc
```diff
@@ -178,37 +181,39 @@
...
-                      tuples-b))]
-        (Relation. attrs-a tuples'))
-
-      :else
-      (let [all-attrs (zipmap (keys (merge attrs-a attrs-b)) (range))]
-        (-> (Relation. all-attrs [])
-            (sum-rel a)
-            (sum-rel b))))))
+(defn sum-rel
+  ([a] a)
```
whilo, line 185: Why is this needed now?

whilo (commented):
Overall this looks good to me, the search changes make sense, but I still need another round of review to go through the query namespace and macros in more detail. Lmk if you would like to get feedback on something specific as well. I need to understand what is happening and you are right that the code is a bit dense and might get harder to pick up, so if we can make it clearer that is also helpful. Nonetheless I think this should already be mergeable, I just want to make sure that we get things right.

src/datahike/db/search.cljc
```diff
@@ -26,28 +24,33 @@
...
   "Checks if database pattern is valid"
-  [pattern]
-  (let [[e a v tx added?] pattern]
+  [pattern can-have-vars]
+  (let [bound-var? (if can-have-vars symbol? (fn [_] false))
+        [e a v tx added?] pattern]
 
     (when-not (or (number? e)
                   (nil? e)
+                  (bound-var? e)
```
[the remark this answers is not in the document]
jonasseglare: Good point! When we call `-search`, it is assumed that the pattern does not contain any symbols, but for `-batch-search` we have to allow for symbols, as well, in order for it to work. That's the reason why `validate-pattern` now has a `can-have-vars` parameter to control whether or not we allow for symbols.

However, I realized that we now only call `validate-pattern` in one place with the `can-have-vars` variable being `true`. So I added calls to `validate-pattern` in a few more places where it make sense, calling it also with `can-have-vars` being `false` in some places, to have better runtime checking for bugs.

src/datahike/query.cljc
jonasseglare: Good point! I don't think we need it.

I believe the reason why it was there is probably that I wanted to use `sum-rel` as a step function for a transducer and that would require this arity. But I probably dropped that approach as it would have made this PR less focused.

src/datahike/query.cljc
```diff
@@ -90,14 +93,11 @@
...
-                     (if (seen key)
-                       [dst seen]
-                       [(conj dst tuple) (conj seen key)])))
-                 [[] #{}]
-                 tuples)))
+  ([tuples]
+   (into [] (distinct-tuples) tuples))
+  ([]
+   (let [seen (HashSet.)]
+     (filter #(.add ^HashSet seen (vec %))))))
```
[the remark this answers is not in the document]
jonasseglare: That's a good point! I wanted this function to be usable also as a transducer constructed from the expression `(distinct-tuples)` and this `(filter ...)` expression returns a filtering transducer that has a stateful predicate. In the two places where we use this transducer we use it eagerly so I don't think we will have any issues because of that.

I thought using a `HashSet` to implement this predicate was an elegant way of doing it. But I now realize it makes the code less portable to ClojureScript. So I can spend some time to see if we can have a pure Clojure solution without Java interop that does not sacrifice performance.

src/datahike/db.cljc
```diff
@@ -467,15 +477,19 @@
...
-         (filter (fn [^Datom d] (contains? filtered-tx-ids (datom-tx d)))))))
+(defn- filter-since
+  ([datoms time-point db] (filter-since datoms time-point db identity))
+  ([datoms time-point db final-xform]
+   (let [since-pred (fn [^Datom d]
+                      (if (date? time-point)
+                        (.after ^Date (.-v d) ^Date time-point)
+                        (>= (.-tx d) time-point)))
+         filtered-tx-ids (dbu/filter-txInstant datoms since-pred db)]
+     (into []
```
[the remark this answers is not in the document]
jonasseglare: That is a good point, but right now I cannot see that it would be a problem.

src/datahike/query.cljc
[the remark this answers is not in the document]
whilo: That would be great!

whilo:
Can you rebase this? I hope it is not too annoying.

src/datahike/query.cljc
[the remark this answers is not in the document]
jonasseglare: I came up with a pure Clojure solution that internally uses `clojure.core/distinct`.

jonasseglare:
Thanks @whilo  for the review so far! I have rebased and also improved the implementation of `distinct-tuples` as you requested.

whilo:
@jonasseglare could you do a rebase again?

jonasseglare:
Before moving forward with this PR, I suggest we first review the code in https://github.com/replikativ/datahike/pull/691 .

jonasseglare:
I rebased on the latest main and force-pushed this branch.

src/datahike/tools.cljc
```diff
@@ -209,3 +234,42 @@
     ([dst x]
      (let [k (f x)]
        (assoc! dst k (conj (get dst k []) x))))))
+
+(defn range-subset-tree
```
whilo, line 238: This function only seems to be used in its test. Do we still need it? It would be good to know what it is for then.

test/datahike/test/query_stats_test.cljc
```diff
@@ -51,14 +41,10 @@
           :rules  {}
           :stats  [{:clause '[?e :age ?a]
                     :rels   [{:bound #{'?a '?e} :rows  6}]
-                    :t      :measurement
-                    :type   :lookup
```
whilo, line 55: It is fine for me to remove this information, in particular if it is constant. I think @jsmassa added it for Datomic compatibility though, but I am not sure.

whilo (changes requested):
Looks good. If we can clarify/resolve the two issues I am fine with merging.

src/datahike/tools.cljc
jonasseglare: It is used in `datahike.query/substitution-expansion` and `datahike.query/make-vec-lookup-ref-replacer`. I think you may have missed it because the diff of `query.cljc` is too big to be displayed by default in Github.

test/datahike/test/query_stats_test.cljc
jonasseglare: Oh, you are right: I believe the `:type :lookup` key-value-pair should still be there, but the the `:lookup-stats :valid` can be removed. I updated `query.cljc` so that we keep `:type :lookup`. This is the behaviour that we used to have in the past, which can be seen by running

```
git diff 4b3a87da refactor-query-and-search -- test/datahike/test/query_stats_test.cljc
```

jonasseglare:
And I did some measurements again, suggesting that these changes will make Datahike about twice as fast:

```
TARGET                      ABS TIME (s)   REL TIME
Some other db                      2.319        10%
Datahike 0.6.1573                 23.727       100%
refactor-query-and-search         12.278        52%
```

src/datahike/tools.cljc
whilo: Thank you for pointing that out!

whilo (approved):
Really nice work. Thank you!

PR #677, opened 2024-06-03
Fix issue 676: make bind-by-fn require all attrs to have values

### LINKED ISSUE #676: [Bug]: bind-by-fn computes wrong result if not all symbols have values

0.6.1659

Ubuntu 22

```
(d/db-with [{:db/id 1, :name  "Ivan",  :age   15}
                           {:db/id 2, :name  "Petr",  :age   22, :height 240, :parent 1}
                           {:db/id 3, :name  "Slava", :age   37, :parent 2}
                           {:db/id 4, :name  "Ivan",  :age   22}])
```

### Describe the bug

When using clauses of type `bind-by-fn`, such as `[(get-else $ ?e :height 300) ?height]`, all symbols that are function arguments must have some values in the context. In this example, we have the symbols `?e` and `$` (the database). If the symbol `?e` is not a constant or part of a relation in the context, the function `datahike.query/bind-by-fn` will silently compute the wrong result.

The function `bind-by-fn` must not produce the wrong result if symbols are missing.

Currently, the query engine evaluates the clauses in the order they are listed in the expression. The following assertion that is part of `datahike.test.query-fns-test/test-query-fns` works because the clause `[?e :age ?age]` will make sure that the context will contain the symbol `?e` before the `get-else` clause is processed:

```clj
(is (= #{[1 15 300] [2 22 240] [3 37 300] [4 22 300]}
             (d/q '[:find ?e ?age ?height
                    :where [?e :age ?age]
                    [(get-else $ ?e :height 300) ?height]] db)))
```

By changing the order of the clauses, the query engine will compute the wrong result and the unit test will fail:

```clj
(is (= #{[1 15 300] [2 22 240] [3 37 300] [4 22 300]}
             (d/q '[:find ?e ?age ?height
                    :where
                    [(get-else $ ?e :height 300) ?height]
                    [?e :age ?age]] db)))
```

jonasseglare:
#### SUMMARY
Fixes #676 using the following changes:

* Rewrite `datahike.query/bind-by-fn` to return nil if not all symbols in the function call can be bound to values.
* Add function `datahike.tools/resolve-clauses`.
* Replace all occurrences in of `(reduce -resolve-clause ...)`  by `(dt/resolve-clauses ...)` to handle the `nil` returned by `datahike.query/bind-by-fn`.
* Add unit test that demonstrates that the bug has been fixed.

*Discussion about the solution to issue 676:*
The code in `bind-by-fn` is currently written in such a way that it assumes all symbols in the function call to be known in the context when the clause is evaluated but there are not checks to ensure that this is actually the case. Maybe there is a way to make it possible for symbols to be unknown but that would probably require a substantial rewrite of `bind-by-fn`. An easier way to address this problem is to make `bind-by-fn` return `nil`, instead of the new context, if not all symbols are known. However, this `nil` will consequently be returned by `-resolve-clause` and needs to be handled. We handle that `nil` by replacing all occurences of `(reduce -resolve-clause ...)` by `(dt/resolve-clauses ...)` to handle the nil and retry failed clauses in subsequent iterations.

##### Bugfix
- [x] Integration tests added
- [ ] Formatting checked

### DIFF
--- a/src/datahike/query.cljc
+++ b/src/datahike/query.cljc
@@ -750,43 +750,55 @@
     (update context :rels conj new-rel)))
 
 (defn bind-by-fn [context clause]
-  (let [[[f & args] out] clause
-        binding (dpi/parse-binding out)
-        fun (or (get built-ins f)
-                (get clj-core-built-ins f)
-                (context-resolve-val context f)
-                (resolve-sym f)
-                (resolve-method f)
-                (when (nil? (rel-with-attr context f))
-                  (dt/raise "Unknown function '" f " in " clause
-                            {:error :query/where, :form clause, :var f})))
-        [context production] (rel-prod-by-attrs context (filter symbol? args))
-        new-rel (if fun
-                  (let [tuple-fn (-call-fn context production fun args)
-                        rels (for [tuple (:tuples production)
-                                   :let [val (tuple-fn tuple)]
-                                   :when (not (nil? val))]
-                               (prod-rel (Relation. (:attrs production) [tuple])
-                                         (in->rel binding val)))]
-                    (if (empty? rels)
-                      (prod-rel production (empty-rel binding))
-                      (reduce sum-rel rels)))
-                  (prod-rel (assoc production :tuples []) (empty-rel binding)))
-        idx->const (reduce-kv (fn [m k v]
-                                (if-let [c (k (:consts context))]
-                                  (assoc m v c)     ;; different value at v for each tuple
-                                  m))
-                              {}
-                              (:attrs new-rel))]
-    (if (empty? (:tuples new-rel))
-      (update context :rels collapse-rels new-rel)
-      (-> context                                        ;; filter output binding
-          (update :rels collapse-rels
-                  (update new-rel :tuples #(filter (fn [tuple]
-                                                     (every? (fn [[ind c]]
-                                                               (= c (get tuple ind)))
-                                                             idx->const))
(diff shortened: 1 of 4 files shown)


src/datahike/query.cljc
```diff
@@ -750,43 +750,55 @@
...
+          [context production] (rel-prod-by-attrs context attrs)
+          symbols-with-values (into #{}
+                                    (mapcat keys)
+                                    [(:attrs production)
+                                     (:consts context)
+                                     (:sources context)])]
+      ;; Currently, we can only evaluate this clause if all variables
+      ;; in the function call are bound. If not, we return nil which
+      ;; is handled by `datahike.tools/resolve-clauses`.
+      (when (every? symbols-with-values attrs)
```
jonasseglare, line 773: This `(when ...)` is the only relevant change in this function. Everything else is pretty much change of indentation and binding `attrs` and `symbols-with-values`.

whilo: That makes sense. I think the indentation changed though, which makes it harder to see changes and also will trigger our format checker.

whilo (changes requested):
Should we throw an exception when not all arguments to a function are bound? @jsmassa and I discussed doing some static analysis (before running the query) to determine whether variables are bound or clause reordering was necessary. Maybe I misunderstand though.

whilo (approved):
Thanks, after our discussion on slack this is good to go!

PR #681, opened 2024-06-05
Fix issue 680: correctly pull attributes

### LINKED ISSUE #680: [Bug]: Pulled attributes are not correct when attribute-refs are used

0.6.1659

Ubuntu 22

```
{:store {:backend :mem, :id "ref-config"},
 :writer {:backend :self},
 :attribute-refs? true,
 :keep-history? false,
 :schema-flexibility :write}
```

### Describe the bug

I added the following code to `datahike.test.attribute-refs.query-pull-test`: 

```clj
(deftest test-pull-attribute
  (testing "Pull an attribute"
    (let [db (d/db-with ref-db [{:db/ident :attribute-to-use
                                 :db/valueType :db.type/keyword}])
          result (d/q '[:find (pull ?attr [*])
                        :in $ ?attr-name
                        :where
                        [?attr :db/ident ?attr-name]]
                      db :attribute-to-use)
   ....
```

When I run this code, the `result` gets bound to the following value:

```clj
[[#:db{:id 62,
       :ident :attribute-to-use,
       :valueType 23,
       :cardinality 11}]]
```

This is not what we want. We would like the values at `:db/valueType` and `:db/cardinality` to be maps that have at least the keys `:db/id` but also `:db/ident`, because that is what Datomic would return.

We want `result` to be something like the following (apart from possibly different values at the `:db/id` keys):

```clj
[[#:db{:id 62,
       :ident :attribute-to-use,
       :valueType #:db{:id 23, :ident :db.type/keyword},
       :cardinality #:db{:id 11, :ident :db.cardinality/one}}]]
```

Add the following unit test to the namespace `datahike.test.attribute-refs.query-pull-test`:

```clj
(deftest test-pull-attribute
  (testing "Pull an attribute"
    (let [db (d/db-with ref-db [{:db/ident :attribute-to-use
                                 :db/valueType :db.type/keyword}])
          result (d/q '[:find (pull ?attr [*])
                        :in $ ?attr-name
                        :where
                        [?attr :db/ident ?attr-name]]
                      db :attribute-to-use)
          [[x]] result
          ids [(:db/id x)
               (-> x :db/valueType :db/id)
               (-> x :db/cardinality :db/id)]]
      (is (= 1 (count result)))
      (is (= 1 (count (first result))))
      (is (= #{:db/id :db/ident :db/valueType :db/cardinality} (set (keys x))))
      (is (number? (:db/id x)))
      (is (= :attribute-to-use (:db/ident x)))
      (is (= :db.type/keyword (-> x :db/valueType :db/ident)))
      (is (= :db.cardinality/one (-> x :db/cardinality :db/ident)))
      (is (every? number? ids))
      (is (= (count (set ids))
             (count ids))))))
```

Currently, this unit test would fail against the main branch. If we fix the problem, it is going to succeed.

jonasseglare:
#### SUMMARY

Fixes #680 

This PR fixes a bug related to pulling attributes. See the issue description and the unit test.

In short, if we run the following query

```clj
(let [db (d/db-with ref-db [{:db/ident :attribute-to-use
                                 :db/valueType :db.type/keyword}])
          result (d/q '[:find (pull ?attr [*])
                        :in $ ?attr-name
                        :where
                        [?attr :db/ident ?attr-name]]
                      db :attribute-to-use)
```

the code in this PR makes sure that we get the **correct result** of 

```clj
[[#:db{:id 62,
       :ident :attribute-to-use,
       :valueType #:db{:id 23, :ident :db.type/keyword},
       :cardinality #:db{:id 11, :ident :db.cardinality/one}}]]
```

instead of the **incorrect result** of

```clj
[[#:db{:id 62,
       :ident :attribute-to-use,
       :valueType 23,
       :cardinality 11}]]
```

##### Bugfix
- [x] Integration tests added
- [ ] Formatting checked

### DIFF
--- a/src/datahike/db/utils.cljc
+++ b/src/datahike/db/utils.cljc
@@ -28,6 +28,10 @@
           :cljs [^boolean ref?]) [db attr]
   (is-attr? db attr :db.type/ref))
 
+(defn #?@(:clj [^Boolean system-attrib-ref?]
+          :cljs [^boolean system-attrib-ref?]) [db attr]
+  (is-attr? db attr :db/systemAttribRef))
+
 (defn #?@(:clj [^Boolean component?]
           :cljs [^boolean component?]) [db attr]
   (is-attr? db attr :db/isComponent))
@@ -152,8 +156,11 @@
       {:ident attr :ref (dbi/-ref-for db attr)})
     {:ident attr :ref attr}))
 
+(defn ident-name? [x]
+  (or (keyword? x) (string? x)))
+
 (defn validate-attr-ident [a-ident at db]
-  (when-not (or (keyword? a-ident) (string? a-ident))
+  (when-not (ident-name? a-ident)
     (raise "Bad entity attribute " a-ident " at " at ", expected keyword or string"
            {:error :transact/syntax, :attribute a-ident, :context at}))
   (when (and (= :write (:schema-flexibility (dbi/-config db)))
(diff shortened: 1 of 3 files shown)


whilo (approved):
Looks good. Thanks!

PR #683, opened 2024-06-05
Fix issue 682: Make history, as-of and since composable with each other

### LINKED ISSUE #682: Make history, as-of and since composable

Currently the following three queries all produce an error:

```clj
(d/q '[:find ?a :in $ ?e :where [?e :age ?a]]
                        (-> @conn d/history d/history)
                        [:name "Alice"])

(d/q '[:find ?a :in $ ?e :where [?e :age ?a]]
                    (-> @conn
                        (d/as-of tx-id0)
                        d/history)
                    [:name "Alice"])

(d/q '[:find ?a :in $ ?e :where [?e :age ?a]]
                    (-> @conn
                        d/history
                        (d/as-of tx-id0))
                    [:name "Alice"])
```

The error is the following:

```
error: java.lang.IllegalArgumentException: No implementation of method: :-slice of protocol: #'datahike.index.interface/IIndex found for class: nil
```

I would like to make queries against database expressions such as `(-> @conn d/history d/history)`, `(-> @conn (d/as-of tx-id0) d/history)` where we compose calls to `d/history`, `d/as-of` and `d/since`.

Datomic, where this is possible (at least to some extent).

jonasseglare:
#### SUMMARY

Fixes #682 

Makes it possible to compose `history`, `as-of` and `since` in arbitrary ways, e.g. `(-> @conn history (as-of tx-id) (since tx-id2))`.

##### Feature
- [x] Integration tests added
- [ ] Documentation added
- [ ] Formatting checked

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -137,6 +137,68 @@
       (update :aevt di/-persistent!)
       (update :avet di/-persistent!)))
 
+(defn contextual-search-fn [{:keys [temporal?]}]
+  (case temporal?
+    true dbs/temporal-search
+    false dbs/search-current-indices))
+
+(defn contextual-search [db pattern context]
+  ((contextual-search-fn context) db pattern))
+
+(defn contextual-datoms [db index-type cs {:keys [temporal?]}]
+  (case temporal?
+    true (dbu/temporal-datoms db index-type cs)
+    false (di/-slice
+           (get db index-type)
+           (dbu/components->pattern db index-type cs e0 tx0)
+           (dbu/components->pattern db index-type cs emax txmax)
+           index-type)))
+
+(defn contextual-seek-datoms [db index-type cs {:keys [temporal?]}]
+  (case temporal?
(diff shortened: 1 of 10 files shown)


src/datahike/db.cljc
```diff
@@ -137,6 +137,68 @@
...
+
+(defn deeper-index-range [origin-db db attr start end context]
+  (dbi/-index-range origin-db
+                    attr
+                    start
+                    end
+                    (merge {:current-db db}
+                           context)))
+
+(defmacro todo [x] x)
```
whilo, line 199: Is this a leftover?

whilo (changes requested):
Overall this looks good. I think not all of these compositions make sense, but it is nice if it works this way.

src/datahike/db.cljc
jonasseglare: Yes, thanks for spotting that. I have removed it.

whilo (approved):
That makes sense, thanks for providing this improved compatibility with Datomic.

PR #687, opened 2024-06-25
Update datalog-parser dependency: Support pull patterns being sets

### LINKED ISSUE #686: Support pull patterns being sets

It would be nice to be able to provide pull patterns being sets, so that we can, for example, write code like this:

```clj
  (is (= (d/q '[:find (pull ?e #{:name}) .
                :where [?e :age 25]]
              test-db)
         {:name "Ivan"}))
```

where the pull pattern is a set: `#{:name}`. This kind of syntax is supported by Datomic, so it would be good if Datahike supports it too.

Support for this was added in the datalog-parser recently: https://github.com/replikativ/datalog-parser/commit/e2c0f3dd8ababbc07bcbd5ce22e1fd1d36b7f85e

We just have to update the datalog-parser dependency:

```clj
io.replikativ/datalog-parser                {:mvn/version "0.2.29"}
```

We could choose to not change the API and force the user to always provide pull patterns that are `sequential?`. But this incompatibility with Datomic might be less convenient for the user.

jonasseglare:
#### SUMMARY

Fixes #686 

Update the datalog parser dependency to version `0.2.29` to allow pull patterns to be *sets*, so that we can write queries like this one:

```clj
(d/q '[:find (pull ?e #{:name}) .
      :where [?e :age 25]]
     test-db)
```

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [ ] Formatting checked

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -7,7 +7,7 @@
                                                      :exclusions [org.clojure/clojurescript]}
         io.replikativ/superv.async                  {:mvn/version "0.3.46"
                                                      :exclusions [org.clojure/clojurescript]}
-        io.replikativ/datalog-parser                {:mvn/version "0.2.28"}
+        io.replikativ/datalog-parser                {:mvn/version "0.2.29"}
         io.replikativ/zufall                        {:mvn/version "0.2.9"}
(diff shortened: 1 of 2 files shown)


whilo (approved):
Thanks!

PR #691, opened 2024-08-28
Refactor db interface

jonasseglare:
This PR mainly refactors the code in `src/datahike/db.cljc` and `src/datahike/db/interface.cljc` so that datom filtering and evaluation of current datoms from historical ones is moved into the single function `datahike.db/post-process-datoms`. This makes the code easier to reason about and deduplicates some calls. These changes will also make it easier to move forward with the code in the PR https://github.com/replikativ/datahike/pull/674 .

I had to introduce a record `SearchContext` to make it fast enough and resort to Java interop syntax to access fields. Otherwise, it is too slow.

I measured the performance change and this code takes rougly 2% less time in my benchmark which is a small improvement:

```
TARGET                          ABS TIME (s)   REL TIME
Official Datahike                     49.765       100%
Refactor db interface                 48.666        98%
```

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -168,7 +168,7 @@
   (let [xform (dbi/context-xform context)
         time-pred (dbi/context-time-pred context)]
     (cond
-      
+
       time-pred
       (let [time-xform (temporal-datom-filter datoms time-pred db)]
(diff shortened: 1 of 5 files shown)


whilo (approved):
Looks good. Thanks!

PR #698, opened 2024-09-24
Use ident keywords in add/retract tuple syntax with attribute-refs? enabled

### LINKED ISSUE #697: Support ident keywords with add/retract tuple syntax when attribute-refs? is enabled

It would be useful if we could run code such as 

```clj
(d/transact conn [[:db/add i :aka "Devil"]])
(d/transact conn [[:db/add i :aka "Tupen"]])
```

even when `attribute-refs?` is set to true. Currently, this is not possible because the `:aka` keyword in the above expressions will cause an error.

Make it possible to run the code above. I wrote a full unit test to be included in the file `test/datahike/test/attribute_refs/transact_test.cljc` once the feature has been implemented.

This is the unit test that reflects how I would like it to work.

```clj
(deftest transact-with-ident-attrs
  (let [conn (setup-new-connection)
        e0 (:max-eid @conn)
        i (inc e0)]
    (d/transact conn [[:db/add i :aka "Devil"]])
    (d/transact conn [[:db/add i :aka "Tupen"]])
    (is (= #{["Devil"] ["Tupen"]}
           (d/q `[:find ?v
                  :where [~i :aka ?v]] @conn)))
    (d/transact conn [[:db/retract i :aka  "Tupen"]])
    (is (= #{["Devil"]}
           (d/q `[:find ?v
                  :where [~i :aka ?v]] @conn)))
    (d/release conn)))
```

Call the API with the actual attribute-refs instead, such as

```clj
(d/transact conn [[:db/add i 345 "Devil"]])
(d/transact conn [[:db/add i 345 "Tupen"]])
```

But Datomic seems to support having keywords as attributes, so it would be good if Datahike does too.

jonasseglare:
Fixes #697

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [ ] Formatting checked


src/datahike/db/utils.cljc
```diff
@@ -146,11 +146,21 @@
   (when eid
     (entid-strict db eid)))
 
+(defn attr-has-ref? [db attr]
+  (and (not (ds/is-system-keyword? attr))
+       (:attribute-refs? (dbi/-config db))
+       (not (nil? attr))))
+
+(defn attr-ref-or-ident [db attr]
+  (if (and (attr-has-ref? db attr)
```
whilo, line 155: Can you swap the two conditionals in the `and` here such that the number case still short circuits? This would save a bit of work in this case.

whilo (changes requested):
Looks good and pretty handy that this is supported now. If you change the clause and rebase I will merge.

src/datahike/db/utils.cljc
jonasseglare: Thanks, I ended up changing order in two places. Good idea!

whilo (approved):
Thank you!

PR #702, opened 2024-09-25
always log to STDERR from datahike.cli

### LINKED ISSUE #694: [Bug]: CLI: some warnings are being output to stdout

0.6.1576 (CLI)

Whatever was used to build `dthk` v0.6.1576

Ubuntu 22.04

```edn
{:keep-history? true
 :store
 {:backend :file
  :path "/home/tamayo/projects/bread-cms/dthk.db"
  :config {:in-place? true}}
 :attribute-refs? true
 :schema-flexibility :write

 ;; These are part of an integration and are included here for completeness:
 :force? true
 :recreate? true
 :db/type :datahike-cli
 :cli/dthk-path "/home/tamayo/bin/dthk"
 :cli/config-path "/home/tamayo/projects/bread-cms/dthk.edn"}
```

### Describe the bug

Some warnings are getting printed to stdout instead of stderr:

```
$ dthk query "[:find ?e . :in $ :where [?e :bad/attr]]" db:dthk.edn 2>/dev/null
2024-09-19T07:18:41.583Z toast WARN [datahike.db:?] - Attribute :bad/attr has not been found in database
60
$ dthk query "[:find ?e . :in $ :where [?e :bad/attr]]" db:dthk.edn >/dev/null
$
```

I expect all warnings/errors to print to stderr, so I can reliably parse results as EDN (and check for warnings separately).

```
$ dthk query "[:find ?e . :in $ :where [?e :bad/attr]]" db:dthk.edn 2>/dev/null
60
$ dthk query "[:find ?e . :in $ :where [?e :bad/attr]]" db:dthk.edn >/dev/null
2024-09-19T07:18:41.583Z toast WARN [datahike.db:?] - Attribute :bad/attr has not been found in database
```

Run the above command with latest CLI :)

acobster:
#### SUMMARY

Fixes #694 

This is an alternative proposal to #699 

##### Bugfix
- [ ] Integration tests added - **happy to add to `run-native-image-tests` if we agree this is a good path forward**
- [ ] Formatting checked

#### ADDITIONAL INFORMATION

While #699 solves the proximate issue, I think that always logging to stderr is actually _more_ flexible.

For one thing, logging to stderr is more Unix-y. That's a good thing because it's fewer knobs to turn in the CLI itself. To log to a file you can just use standard redirection:

```sh
dthk query '[:find ...]' 2>/path/to/log/file
```

Also, requiring a logfile precludes the possibility of logging to stderr (or stdout for that matter), which you might want to do. One big advantage of doing so is capturing the output via `clojure.java.shell/sh`:

```clj
(let [{:keys [out err]} (sh "/path/to/dthk" "query" "[:find ...]")
  (when err (log/warn (str "Datahike error: " err))
  ,,,)
```

This is what I intend to do [in Bread's CGI binary](https://github.com/breadsystems/bread-cms/blob/feature/cgi/plugins/datahike/systems/bread/alpha/plugin/datahike_cli.cljc#L27). With stderr logging, I can detect when `dthk` complains about something, and bubble that up as a warning from Bread's code. Without it, I would need to add another knob in Bread to configure a separate Datahike logfile (so it can call `dthk --log-file ...` with the correct path), and it would still be up to users to watch said logfile themselves.

### DIFF
--- a/src/datahike/cli.clj
+++ b/src/datahike/cli.clj
@@ -16,6 +16,13 @@
 
 ;; This file is following https://github.com/clojure/tools.cli
 
+(log/merge-config!
+  {:appenders {:println {:doc "Always prints to *err*"
+                         :enabled? true
+                         :fn (fn log-to-stderr [{:keys [output_]}]
+                               (binding [*out* *err*]
+                                 (println (force output_))))}}})
+
 (defn usage [options-summary]
   (->> [datahike-logo
         "This is the Datahike command line interface."


acobster:
@whilo @TimoKramer pinging you here for visibility. :) Happy to be wrong about the above, just wanted to share some perspective. Thanks!

whilo (approved):

whilo:
Thank you! I agree that redirecting to STDERR for now should do the trick.

PR #706, opened 2024-09-26
Make sure that the datoms of history are distinct

### LINKED ISSUE #705: [Bug]: Datoms from history of empty database with attribute-refs? being true are not distinct

0.6.1584

openjdk 21.0.4 2024-07-16 LTS

Mac OS X

{:store {:backend :mem}
 :keep-history? true
 :attribute-refs? true
 :schema-flexibility :write}

### Describe the bug

If I call

```clj
(-> @conn
     d/history
     (d/datoms {:index :aevt
                :components [:db/txInstant]
                :limit -1})
```
on an empty database, it returns

```clj
[#datahike/Datom [536870912
                  9
                  #inst "1970-01-01T00:00:00.000-00:00"
                  536870912]
 #datahike/Datom [536870912
                  9
                  #inst "1970-01-01T00:00:00.000-00:00"
                  536870912]]
```

That is, it returns **duplicate** datoms. We would want only one datom.

That it returns only one distinct datom in the previous example, that is

```clj
[#datahike/Datom [536870912
                  9
                  #inst "1970-01-01T00:00:00.000-00:00"
                  536870912]]
```

Add this unit test to `datahike.test.attribute-refs.datoms-test`:

```clj
(deftest test-distinct-datoms-history-db
  (with-connect [conn (-> cfg
                          provide-unique-id
                          recreate-database)]
    (is (= 1 (-> @conn
                 d/history
                 (d/datoms {:index :aevt
                            :components [:db/txInstant]
                            :limit -1})
                 count)))))
```

It will fail, which illustrates the bug.

jonasseglare:
#### SUMMARY
Make sure that datoms returned from the history of a database are distinct.

Fixes #705 

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked

##### Feature
- [ ] Integration tests added
- [ ] Documentation added
- [ ] Formatting checked

### DIFF
--- a/src/datahike/db/utils.cljc
+++ b/src/datahike/db/utils.cljc
@@ -204,10 +204,15 @@
 
 (defn distinct-datoms [db current-datoms history-datoms]
   (if  (dbi/-keep-history? db)
-    (concat (filter #(or (no-history? db (:a %))
-                         (multival? db (:a %)))
-                    current-datoms)
-            history-datoms)
+    (let [current-datoms (filterv (fn [datom]
+                                    (let [a (:a datom)]
+                                      (or (no-history? db a)
+                                          (multival? db a))))
+                                  current-datoms)
+          current-datom-set (set current-datoms)]
+      (into current-datoms
+            (remove current-datom-set)
+            history-datoms))
     current-datoms))
 
 (defn temporal-datoms [db index-type cs]
--- a/test/datahike/test/api_test.cljc
+++ b/test/datahike/test/api_test.cljc
@@ -874,3 +874,18 @@
 
 (deftest test-metrics-attr-refs
   (test-metrics (assoc metrics-base-cfg :attribute-refs? true)))
+
+(deftest test-no-history-datoms-in-empty-db
+  (testing "When attribute-refs? is false, there are no initial datoms"
+    (utils/with-connect [conn (-> {:store {:backend :mem
+                                           :id "hashing"}
+                                   :keep-history? true
+                                   :schema-flexibility :write}
+                                  utils/provide-unique-id
+                                  utils/recreate-database)]
+      (is (-> @conn
+              d/history
+              (d/datoms {:index :aevt
+                         :components []
+                         :limit -1})
+              empty?)))))
--- a/test/datahike/test/attribute_refs/datoms_test.cljc
+++ b/test/datahike/test/attribute_refs/datoms_test.cljc
@@ -108,3 +108,33 @@
                               (:e datom)
                               (:v datom)
                               (:tx datom)]})))))))
+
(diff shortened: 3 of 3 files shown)


src/datahike/db/utils.cljc
```diff
@@ -204,10 +204,15 @@
...
-                         (multival? db (:a %)))
-                    current-datoms)
-            history-datoms)
+    (let [current-datoms (filterv (fn [datom]
+                                    (let [a (:a datom)]
+                                      (or (no-history? db a)
+                                          (multival? db a))))
+                                  current-datoms)
+          current-datom-set (set current-datoms)]
+      (into current-datoms
```
whilo, line 213: If I understand this correctly this transducer acts on `current-datoms` which is a persistent vector. This would mean you have to eagerly realize the full datom result set, while before this was lazy. Is this correct? It would be good to keep operations that act on large chunks of the database lazy, I think.

whilo (changes requested):
Overall looks good, will merge once we addressed the laziness question.

src/datahike/db/utils.cljc
jonasseglare: I pushed some code but I am working on a better solution.

jonasseglare: OK, I tried to implement an approach where we assume that the two datoms sequences to merge are sorted and then try to merge the sorted sequences. If that would have worked, we would have a lazy sequence with distinct datoms without having to store all the current datoms in memory at a given time. An attempt is found here:

https://github.com/replikativ/datahike/compare/main...jonasseglare:datahike:distinct-history-datoms-dev?expand=1

But I was not able to get it to work. So I think that the code that as pushed is the laziest I can come up with for now.

jonasseglare:
See my comment in the code. I was able to make it lazier but it is not completely lazy.

src/datahike/db/utils.cljc
whilo: I like the merge function, I have thought about this approach in the past as well. Why does it not work?

whilo:
I think this is good enough, the one disadvantage is that there is cliff now in the lazy seq once the set needs to be computed, but this computation is at least necessary to do the following job.

jonasseglare:
I gave the approach of merging sorted distinct sequences a second chance and was able to get it to work.

However, I did notice a test breaking but it might potentially be due to a bug in `load-entities`, see the changes in the file [`migrate_test.clj`](https://github.com/replikativ/datahike/pull/706/files#diff-320a47b53eedb3d8ba71aa30bf25d023ebeebf6514951651fd20b155c71a9e95)

whilo (approved):
Nice work, thanks!

PR #708, opened 2024-09-26
Use squential UUID for store keys.

### LINKED ISSUE #707: Improve memory locality

Currently datahike uses random UUIDs for the underlying store, which scatters written entries in its key space.

Using squential UUIDs will locate things together that are written together, which often is a good proxy for memory locality.

-

whilo:
Fixes #707


whilo:
I had forgotten to add the specification test case for tempid yesterday, I just added this here as well.

jonasseglare (approved):
Looks good! So the important change here is that persistent set uses `squuid` instead of just `uuid`.

whilo:
I realized this actually needs to be fixed here, since `uuid` is again called on the key for the store there, https://github.com/replikativ/konserve/blob/main/src/konserve/impl/defaults.cljc#L37.

PR #713, opened 2024-09-27
Fix as-of and since so that they include and exclude time-point respectively

### LINKED ISSUE #712: [Bug]: as-of does not include time-point and since does not exlude time-point

0.6.1586

openjdk 21.0.4 2024-07-16 LTS

Mac OS X

Doesn't matter.

### Describe the bug

I assume that the functions `as-of` and `since` should work just like they work in Datomic? From the Datomic docs, it says

For `as-of`:
> An as-of filter returns a database "as of" at a particular point in time, ignoring any transactions after that point.

For `since`:
> A since filter is the opposite of as-of. Taking the same point-in-time arguments as as-of, since returns a value of the database that includes only datoms added by transactions after that point in time.

From the above, I read that `as-of` will *include* the time point, whereas `since` will *exclude* the time point.

But this is not how it is implemented in Datahike, in case `time-point` is a `Date`. I believe this is a mistake:
```clj
(defn- as-of-pred [time-point]
  (if (date? time-point)
    (fn [^Datom d] (.before ^Date (.-v d) ^Date time-point))
    (fn [^Datom d] (<= (dd/datom-tx d) time-point))))

(defn- since-pred [time-point]
  (if (date? time-point)
    (fn [^Datom d] (.after ^Date (.-v d) ^Date time-point))
    (fn [^Datom d] (>= (.-tx d) time-point))))
```

That `as-of` will include the time point and `since` will exclude it.

On my Macbook, I discovered a unit test that uses `as-of`. It only fails sometimes, namely when the `time-point` is *the same* as the timestamp of the last transaction. Then I get this error:

```
ERROR in datahike.test.time-variance-test/test-as-of-db (utils.cljc:142)
get values at specific time
Exception: clojure.lang.ExceptionInfo: Nothing found for entity id [:name "Alice"]
{:error :entity-id/missing, :entity-id [:name "Alice"]}
```

jonasseglare:
#### SUMMARY

Implement `as-of` and `since` so that they work just like in Datomic.

* Implement `as-of` so that it *includes* time-point
* Implement `since` so that it *excludes* time-point

Fixes #712 

##### Bugfix
- [x] Integration tests added
- [ ] Formatting checked

### DIFF
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -129,15 +129,20 @@
   #?(:cljs (instance? js/Date d)
      :clj (instance? Date d)))
 
-(defn- as-of-pred [time-point]
+(defn- date<= [^Date a ^Date b]
+  (not (.before b a)))
+
+(defn- as-of-pred
+  "Create an as-of predicate, *including* the time-point"
+  [time-point]
   (if (date? time-point)
-    (fn [^Datom d] (.before ^Date (.-v d) ^Date time-point))
+    (fn [^Datom d] (date<= (.-v d) time-point))
     (fn [^Datom d] (<= (dd/datom-tx d) time-point))))
 
-(defn- since-pred [time-point]
-  (if (date? time-point)
-    (fn [^Datom d] (.after ^Date (.-v d) ^Date time-point))
-    (fn [^Datom d] (>= (.-tx d) time-point))))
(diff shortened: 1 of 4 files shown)


whilo (approved):
Thanks again! :)

PR #727, opened 2025-03-14
Updated import db

### LINKED ISSUE #633: [Bug]: schema/double changed from 0.4 to 0.6

0.6

17

guix

irrelevant to this ticket

### Describe the bug

I started my datahike database with version 0.4 with schema-on-write.
I ONLY used double in the schema definition.
Now I had to migrate the 0.4 edn dumps to import in 0.6 which uses cbor dumps.
Previously I was exporting in 0.4 edn dumps which I would import on another machine without a problem.
Now I have noted that when the edn dumps are read via the edn parser and imported into 0.6
that I had to make an adjustment to change all float values to double values.
Interestingly, this was not a problem before.

put into readme that :double has a different meaning in 0.4 in relationship to 0.6.

```
(defn float->double [v]
  (if (float? v)
    (double v)
    v))

(defn db-migrate []
  (warn "migrating cbor db..")
  (let [txs-old (load-cbor crbdb/conn "data/datahike-dump/eavt-dump")
        s (stats txs-old)
        max-eid (:max-eid s)
        max-tx (:max-tx s)
        tx-old-no-schema (remove datom-schema? txs-old)
        tx-old-safe (map #(-> %
                              (update :v float->double))
                         tx-old-no-schema)
        schema-with-eids (assoc-schema-ids max-eid schema)]
    (warn "dump stats: " s)
    (warn "transacting schema with eid above: " max-eid)
    (let [result-schema (crbdb/transact schema-with-eids)]
      (print-tx-stats result-schema))
    (let [result-import (api/transact crbdb/conn (vec tx-old-safe))]
      (warn "import dump result:")
      (print-tx-stats result-import))
    ;(assoc crbdb/conn :max-tx max-tx)
    (warn "db migration finished!")))
```

alekcz:
#### SUMMARY
The current `import-db` function does not work at scale. It requires a lot of memory and takes very long. 
Restoring our production instance took 2 hours for 16M datoms and required a huge instance. This is not practical. 

This PR changes `import-db` such that one thread reads the serialized data token by token while another loads them into the store. This has changed the restore of 16M datoms from 2 hours to 21 seconds. 

This introduces 3 new options.

```clojure
(defn import-db [conn path & opts] ...)
```

**filter-schema?**:   
ignore schema datoms when importing the db. This is helpful if your schema has evolved over time. 

**load-entities?**:   
uses load-entities instead of transact. This is much faster. It is also more forgiving where intial datoms do not adhere to current schema

**sync?**:  
When set to true old behaviour is preserved, when set to false importing is async in two threads. A status function is return to allow querying of the status

```clojure
=> (status)
{:complete? false
 :preprocessed? true
 :total-datoms 1200000
 :remaining 200000})
```

For backwards compatibility the default mode is synchronous, uses transact, and does not filter the schema. 

##### Bugfix
- [x] Related issues linked using `fixes #633`

##### Feature
- [x] Related issues linked using `fixes #633`
- [ ] Integration tests added
- [ ] Documentation added
- [x] Formatting checked

#### ADDITIONAL INFORMATION

I would appreciate a second pair of eyes on this. Here's how you can run your own test

```clojure
(comment
  ;; include
  ;; [io.replikativ/datahike-jdbc "0.3.49"]
  ;; [org.xerial/sqlite-jdbc "3.41.2.2"]
  (require '[stub-server.migrate :as m] :reload)
  (require '[datahike.api :as d])
  (require '[clojure.java.io :as io])
  (require '[datahike-jdbc.core])
  (def dev-db (str "./temp/sqlite/ingest"))
  (io/make-parents dev-db)
  ;; for testing large db's it's best to use sqlite. File store may run out of resources threads or pointers 
  ;; my test 21 seconds with SQLite with 16M datoms
  (def cfg      {:store {:backend :jdbc :dbtype "sqlite" :dbname dev-db}
                 :keep-history? true
                 :allow-unsafe-config true
                 :store-cache-size 20000
                 :search-cache-size 20000})
  (d/create-database cfg)
  (def conn (d/connect cfg))
  (def status (m/import-db conn "prod.backup.cbor" {:sync? true :filter-schema? false})))
```

### DIFF
--- a/src/datahike/migrate.clj
+++ b/src/datahike/migrate.clj
@@ -3,7 +3,13 @@
             [datahike.constants :as c]
             [datahike.datom :as d]
             [datahike.db :as db]
-            [clj-cbor.core :as cbor]))
+            [clj-cbor.core :as cbor]
+            [clojure.core.async :as async]
+            [clojure.java.io :as io]
+            [taoensso.timbre :as timbre]
+            [clj-cbor.tags.time :as tags.time])
+  (:import [java.util.concurrent LinkedBlockingQueue]
+           [java.sql Timestamp]))
 
 (defn export-db
   "Export the database in a flat-file of datoms at path.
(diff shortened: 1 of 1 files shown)


alekcz:
@whilo @TimoKramer @awb99 @yflim a review would be greatly appreciated 🙏

src/datahike/migrate.clj
```diff
@@ -17,23 +23,127 @@
...
+        (when (not @stop) (recur))))
+
+      ;; allow the user to stop or monitor the ingestion thread with this atom
+    (fn []
+      {:complete? @stop
+       :preprocessed? @processed
+       :total-datoms @datom-count
+       :remaining (- @datom-count @txn-count)})))
+
+(comment
```
whilo, line 130: @alekcz Thank you so much for helping with this! Overall it looks good to me and it is great that this code performs well for you. However, I cannot evaluate this comment block, stub-server is not in scope and the alias d is already loaded. Could you turn this into a test?

whilo (changes requested):
@alekcz see comment.

whilo:
I read through this again and I am fine with merging it even without the test, since it is already covered and definitely better than before. I am wondering whether it is right to export the Datoms sorted by eid though, I would have guessed it would be better to sort by tx-id. Also, sorting is slow. @alekcz Do you use export-db even?

Issue #739, opened 2025-07-20
[Bug]: Upsert fails for tupleAttr identity


jessie-ross:
### What version of Datahike are you using?

0.6.1599 and main branch

### What version of Java are you using?

OpenJDK 64-Bit Server VM 21.0.4+7-LTS

### What operating system are you using?

MacOS

### What database EDN configuration are you using?

{:store {:backend :file :path db-path}}

### Describe the bug

Upserting a `tupleAttr` entity fails when the entity exists.

### What is the expected behaviour?

This should work without throwing:
```clojure
  (d/transact conn
                [
                 {:db/ident :x
                  :db/valueType :db.type/string
                  :db/cardinality :db.cardinality/one}
                 {:db/ident :y
                  :db/valueType :db.type/string
                  :db/cardinality :db.cardinality/one}
                 {:db/ident :x+y
                  :db/valueType :db.type/tuple
                  :db/tupleAttrs [:x :y]
                  :db/cardinality :db.cardinality/one
                  :db/unique :db.unique/identity}])
  (d/transact conn [{:x "asdf" :y "1234"}]))
  (d/transact conn [{:x "asdf" :y "1234"}])))
```

### How can the behaviour be reproduced?

```clojure
;; Plain identity upserts work as expected
(let [conn (-> (d/create-database) (d/connect))]
  (d/transact conn
                [{:db/ident :x
                  :db/valueType :db.type/string
                  :db/cardinality :db.cardinality/one
                  :db/unique :db.unique/identity}])
  (clojure.pprint/pprint (d/transact conn [{:x "asdf"}]))
  (clojure.pprint/pprint (d/transact conn [{:x "asdf"}]))
  (clojure.pprint/pprint (d/transact conn [{:x "asdf"}]))
  (d/q '[:find ?e ?a ?v ?t
         :where [?e ?a ?v ?t]]
       (d/history @conn)))

;; Tuple upserts fails
(let [conn (-> (d/create-database) (d/connect))]
  (d/transact conn
                [
                 {:db/ident :x
                  :db/valueType :db.type/string
                  :db/cardinality :db.cardinality/one}
                 {:db/ident :y
                  :db/valueType :db.type/string
                  :db/cardinality :db.cardinality/one}
                 {:db/ident :x+y
                  :db/valueType :db.type/tuple
                  :db/tupleAttrs [:x :y]
                  :db/cardinality :db.cardinality/one
                  :db/unique :db.unique/identity}])
[...]

PR #740, opened 2025-07-21
Fix upsert-eid to support composite-tuples

jessie-ross:
#### SUMMARY

- Fixes https://github.com/replikativ/datahike/issues/739

Allowing composite tuples to upsert correctly.

I've also added a `flake.nix` for nix convenience, but happy to remove if unwanted.

##### Bugfix
- [x] Formatting checked

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1752950548,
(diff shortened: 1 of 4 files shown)


whilo (approved):
Hey @jessie-ross, thank you for sorting this out! Apologies for the late review. It looks good overall. What kind of tests could we add for tempids and lookup-refs?

PR #752, opened 2025-09-19
ci: update deps and jdk 25 [skip ci]

TimoKramer:
- build with jdk 25
- update build dependencies
- simplify cpp test
- update all mentions of datahike in pod-registry


TimoKramer:
Right, I tested it in my fork and it went well but there are sometimes things that come up.

PR #758, opened 2026-01-14
ci: update base image with latest babashka

TimoKramer:
#### SUMMARY
Changed the CI to use our [self-built container image](https://github.com/replikativ/circleci-orb/pkgs/container/circleci-orb). That image runs now the latest version of babashka, which was previously very old and seems to have made problems. The new image builds on top of [circleci's own clojure image](https://circleci.com/developer/images/image/cimg/clojure#browsers) which includes jdk 21, clojure 1.12, node and chromium browser as well as now with the latest babashka.

### DIFF
--- a/config.edn
+++ b/config.edn
@@ -1,13 +1,13 @@
-{:org "timokramer"
+{:org "replikativ"
  :lib "datahike"
  :version {:major 0
            :minor 7}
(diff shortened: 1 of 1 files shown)


whilo (approved):
Looks good to me. Thanks!

PR #798, opened 2026-03-31
Fix issue #797: Check loop stop condition at beginning of iteration in transact-entities-directly

### LINKED ISSUE #797: [Bug]: transact-entities-directly

0.7.1663

Java(TM) SE Runtime Environment Oracle GraalVM 25+37.1 (build 25+37-LTS-jvmci-b01)

MacOS

Any database will work, I think.

### Describe the bug

When initializing a database using for example `load-entities`, the function `transact-entities-directly` gets called which will loop over all the entities to be transacted. Once it has looped over all entities, it prints out a warning such as the following:

```
2026-03-31T13:16:47.759712Z :warn datahike.db :datahike/attribute-not-found 
  data: {:attribute nil}
```

This warning comes from the fact that inside `transact-entities-directly`, the code starts to process the next entity even if the loop variable `es` of entities to process is empty. Taking the next element from the empty collection results in that element being nil and the code attempts to call `-ref-for` on that element which results in the warning. Only *after this* and a few other operations does it check if the stop condition of the loop.

No warning is printed out at the end of the loop.

Call `load-entities` with some data.

jonasseglare:
Fixes #797

Move the check of the loop stop condition *to the beginning* of each iteration.

### DIFF
--- a/src/datahike/db/transaction.cljc
+++ b/src/datahike/db/transaction.cljc
@@ -875,63 +875,62 @@
 (defn transact-entities-directly [initial-report initial-es]
   (loop [report (update initial-report :db-after transient)
          es initial-es
-         migration-state (or (get-in initial-report [:db-before :migration]) {})]
-    (let [[entity & entities] es
-          {:keys [config] :as db} (:db-after report)
-          [e a v t op] entity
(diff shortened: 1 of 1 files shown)


whilo (approved):
LGTM

Issue #799, opened 2026-03-31
[Bug]: Useless warnings are printed when using pull API


jonasseglare:
### What version of Datahike are you using?

0.7.1663

### What version of Java are you using?

Java(TM) SE Runtime Environment Oracle GraalVM 25+37.1 (build 25+37-LTS-jvmci-b01)

### What operating system are you using?

MacOS

### What database EDN configuration are you using?

{:attribute-refs? true,
   :index :datahike.index/persistent-set,
   :keep-history true,
   :name "jobtech-taxonomy-db",
   :schema-flexibility :write,
   :store {:backend :memory
           :id #uuid "7dc50266-5d1c-beef-a6f8-badcafe10113"}}

### Describe the bug

When using the pull API, e.g. 

```
                     (dc/q '[:find (pull ?edge [:db/id
                                                :mapping.edge/source
                                                :mapping.edge/target
                                                :mapping.edge/mapping])
                             :keys edge
                             :in $ [[?source ?target] ...]
                             :where
                             [?edge :mapping.edge/source ?source]
                             [?edge :mapping.edge/target ?target]]
                           db source-id+target-id-set)
```

lots of warnings are printed. The warnings come from the function `datahike.pull-api/db-ident-and-id` which in turn indirectly calls the function `datahike.db.interface/-ident-for`. The function `datahike.pull-api/db-ident-and-id` is clearly written in such a way that it expects the ident to sometimes be `nil`, but `datahike.db.interface/-ident-for` will always warn if it returns `nil` even if that `nil` is expected.

As a consequence, the call to the pull API tends to result in a lot of false-alarm warnings being spammed, e.g.

```
2026-03-31T13:16:48.081731Z :warn datahike.db :datahike/attribute-ref-not-found 
  data: {:ref 167540}
2026-03-31T13:16:48.081953Z :warn datahike.db :datahike/attribute-ref-not-found 
  data: {:ref 103119}
2026-03-31T13:16:48.082024Z :warn datahike.db :datahike/attribute-ref-not-found 
  data: {:ref 234739}
2026-03-31T13:16:48.082070Z :warn datahike.db :datahike/attribute-ref-not-found 
  data: {:ref 236291}
2026-03-31T13:16:48.082151Z :warn datahike.db :datahike/attribute-ref-not-found 
  data: {:ref 100736}
2026-03-31T13:16:48.082198Z :warn datahike.db :datahike/attribute-ref-not-found 
  data: {:ref 16770}
[...]

PR #800, opened 2026-03-31
Improve logic for missing idents when calling -ident-for

jonasseglare:
#### SUMMARY

Fixes issue #799 

This PR fixes the logic around `datahike.db.interface/-ident-for` such that handling of missing values is more explicit, in the following ways:

* Remove warning of `nil` from `-ident-for` protocol method.
* Add the *function* `ident-for` which accepts a third parameter `missing-strategy` that can be either `:error-on-missing`, `:warn-on-missing` or `:allow-missing`.
* Extend `attr-info` to accept the parameter `missing-strategy` and forward it to `ident-for`.
* Update the calls from `-ident-for` to call `ident-for` with the parameter for the missing strategy. Defensively us `:error-on-missing` whenever possible (as long as unit tests don't break).
* Update the calls of `attr-info` to also specify the `missing-strategy` parameter: Use `:error-on-missing` whenever possible.

As a consequence of these fixes, we get rid of the useless warnings from the pull API as reported in issue #799.

##### Bugfix
- [ ] Integration tests added
- [ ] Formatting checked

### DIFF
--- a/src/datahike/db/interface.cljc
+++ b/src/datahike/db/interface.cljc
@@ -1,4 +1,5 @@
-(ns datahike.db.interface)
+(ns datahike.db.interface
+  (:require [replikativ.logging :as log]))
 
 ;; Database Protocols
 
@@ -117,3 +118,16 @@
 (defprotocol IHistory
   (-time-point [db])
   (-origin [db]))
+
+(defn ident-for [db a-ref missing-strategy]
+  (or (-ident-for db a-ref)
+      (case missing-strategy
+        :error-on-missing (throw (ex-info "Attribute ref not found"
+                                          {:ref a-ref}))
+        ;; TODO: Occurrences of calls with :warn-on-missing should be
+        ;; considered a code smell and an indication that the logic at
+        ;; the call site is not clear and needs to be improved.
+        :warn-on-missing (do (log/warn :datahike/attribute-ref-not-found
+                                       {:ref a-ref})
+                             nil)
+        :allow-missing nil)))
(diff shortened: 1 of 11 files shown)


src/datahike/db/transaction.cljc
```diff
@@ -626,7 +626,8 @@
     (if-let [e (dbu/entid db e)]
       (let [e-datoms (vec (dbi/search db [e]))
             v-datoms (->> (dbi/-attrs-by db :db.type/ref)
-                          (map (partial dbi/-ident-for db))
+                          ;; TODO: How to best handle nil here?
```
whilo: Good question, I would also have to think about it more. What options do you see?

whilo (commented):
Overall looks good to me, but maybe you can rebase and clarify the nil handling case. It is also the only one that has the `:warn-on-missing` case that is unclear. I will have to think about it.

src/datahike/db/interface.cljc
```diff
@@ -117,3 +118,16 @@
 (defprotocol IHistory
   (-time-point [db])
   (-origin [db]))
+
+(defn ident-for [db a-ref missing-strategy]
+  (or (-ident-for db a-ref)
+      (case missing-strategy
+        :error-on-missing (throw (ex-info "Attribute ref not found"
+                                          {:ref a-ref}))
+        ;; TODO: Occurrences of calls with :warn-on-missing should be
```
whilo: Do we still need this? It does not seem to be used anywhere (?).

whilo (changes requested):
Looks good to me in general, can you clarify on the role of the warning?

jonasseglare:
Thanks @whilo for your review! I have addressed your comment about the warning.

whilo (approved):
Thanks!

PR #836, opened 2026-06-01
diff-buf write-buffering + index-root fusion (PSS-backed indexes)

### LINKED ISSUE #143: index selection does not work

whilo:
## diff-buf write-buffering + index-root fusion (PSS-backed indexes)

Two related write-amplification reductions for content-addressed (konserve) persistence, both landing on the persistent-sorted-set index.

### 1. diff-buf (per-child diff buffering at the serialization boundary)
Backed by persistent-sorted-set PR replikativ/persistent-sorted-set#6. A normal commit rewrites the whole root→leaf spine (~`depth+1` object PUTs). With diff-buf, a rewritten branch *buffers* each content-only child's diff into a per-child slot and re-points the child to its durable anchor, so a content-only commit costs **~1 PUT** instead of `~depth+1`. Reads project the buffered diff back lazily on descent (baseline read cost). datahike wiring:
- `:slots` round-trip in the fressian `Branch` read/write handlers (clj + cljs).
- `:diff-buf-size` / `:branching-factor` are create-time-fixed via `:index-config` (round-trips with the store; `connect` adopts the stored value). **Default ON (256) for new stores.**
- crypto-hash / merkle: a branch's content UUID folds its slots; GC/`markFreed` defer freeing to commit so a re-pointed anchor is never freed.

### 2. index-root fusion
Inlines each index's root node into the db-record, so `commit!` skips writing those roots as separate objects (one fewer PUT/commit, and it composes with diff-buf toward ~2 PUTs/commit). **Opt-in, default OFF** (`*default-fuse-index-roots?*`) for now — flipping the default churns object-count assertions across the suite; the SaaS template opts in, and `connect` adopts the stored value so fused and non-fused stores both reconnect. Design note: `doc/index-root-fusion.md`.

### Comparator on the tree (storage stays comparator-agnostic)
The per-index comparator now lives on the PSS and propagates to its Branch nodes (`Branch._projCmp`), instead of being carried on storage. This PR removes `CachedStorage`'s `cmp` field, its `comparator` impl, and `with-comparator`. (Mirrors the upstream PSS change; resolves the earlier review note about a storage-carried comparator.)

### Dependencies (no local checkouts)
- `persistent-sorted-set` → git `a36ecbe…` (branch `feature/op-buf-v5`). Its Java is compiled by tools.deps prep — run **`clojure -X:deps prep`** once after fetching (CI step).
- `konserve` → released **`0.9.349`** (includes the cljs cross-host header meta-size fix #143), replacing the former local dev checkout.

### Validation
- `clj-pss` (default index = persistent-set, diff-buf on): **522 tests / 2475 assertions / 0 failures** (`-Xmx4g`).
- Crash-safety: crash-injection at both sides of the commit HEAD-flip (file backend, diff-buf on) recovers to the old-or-new consistent state — indices agree, no torn/duplicate datoms.
- New regression: `store-test/test-diff-buf-upsert-reopen` (value-changing upserts survive store→reopen with no stale/duplicate datoms — guards the comparator-agnostic `{:absent :present}` leaf-diff serialization).

### Known, non-critical (deferred — not addressed here)
A **fatal `Error`** (e.g. OOM) thrown inside the async write/commit pipeline can hang a synchronous `transact` (the no-timeout deref never completes, because `superv.async` `go-try` catches `Exception`, not `Throwable`). It does **not** corrupt data (commit is copy-on-write + atomic HEAD flip + free-after-flip). Pre-existing, orthogonal to this PR; the clean fix is a deref timeout. Tracked internally.

### Notes
- PSS naming was `op-buf` (hitchhiker/Bε operation buffer); renamed `diff-buf` since this buffers a per-child *diff* at the serialization boundary. History can be squashed on merge.

### DIFF
--- a/build.clj
+++ b/build.clj
@@ -1,6 +1,8 @@
 (ns build
   (:refer-clojure :exclude [compile])
   (:require [clojure.edn :as edn]
+            [clojure.java.io :as io]
+            [clojure.string :as str]
             [clojure.tools.build.api :as b]))
 
 (def class-dir "target/classes")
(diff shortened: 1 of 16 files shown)


whilo:
Superseded by #867 — re-land against current main with persistent-sorted-set's released diff-buf (0.4.130). Most of this PR's serialization work is now upstream in PSS; see #867 for the disposition. Closing.

PR #844, opened 2026-06-26
Query planner on by default + ship src-secondary in jar

whilo:
## Summary

Two related release-hardening changes for the query engine and the maven jar.

### 1. Query planner ON by default + flag rename

The optimizing query planner is now the **default** engine. The base/relational
engine remains a **permanent fallback** for query shapes the planner declines — it
is not deprecated, so the old `*force-legacy*` name was a misnomer.

- `datahike.query/*force-legacy*` → **`*disable-planner*`** (same polarity).
  Default is planner-ON; set env `DATAHIKE_QUERY_PLANNER=false` to disable.
- Fixes the two planner-on failures the planner-OFF CI default had masked:
  - **query-rules-test** (rule var isolation): a disjoint cartesian component that
    projects no user find-var ran with `:find []` and returned the empty set,
    collapsing the whole result. Such an existence-only component now gets one of
    its own free vars (dropped by the final projection) so it yields a row iff
    satisfiable.
  - **query-fns-test / issue-180** (predicate passed via `:in`, used in a
    cross-component post-filter): `eval-post-filter` now short-circuits on empty
    tuples and resolves a var-predicate from the `:in` consts instead of throwing.

### 2. Ship `src-secondary` in the main jar

`config.edn :build :clj :src-dirs` now includes `src-secondary`, so secondary-index
backends ship as source in the jar. Maven users get them without extra deps pulled
onto them (hitchhiker-tree + src-kabel already shipped this way and stay).

## CI

Adds a parallel **`persistent-set-test-base-engine`** job that runs `bb test clj-pss`
with `DATAHIKE_QUERY_PLANNER=false`, so **both engines run on every build**. This is
the gap that let the planner regressions slip in — the prior CI only exercised the
base engine.

## Test results

Full suite green **both settings**:
- planner ON: 1883 tests / 8332 assertions, 0 failures
- planner OFF: 1883 tests / 8305 assertions, 0 failures


whilo:
Pushed a series of fixes validating "planner on by default" against the most important downstream user (jobtech-taxonomy-api: heavy as-of/version queries + recursive rules), plus the native-image release blocker.

**Planner correctness/perf on temporal + recursive workloads** (jobtech unit suite ON now ≡ OFF — 2360 assertions, 0 errors, same single pre-existing failure — and ~10% faster wall-clock):
- `e47c546a` correct fused temporal entity-group execution (as-of/since/historical): AVET-indexing seek gate, all-clause-var projection, card-many optional default
- `a41a9f65` magic-set recursion: `base-scan-attr` tightened to a true linear single-ref-edge (was a `ClassCastException` on value-join bases), temporal-aware `magic-base-scan`/`delta-driven-expand` (was `:-slice of nil`), `magic-base-scan-general` (demand-restricted base, any shape)
- `43fd7594` temporal single pattern-scans take the fused seek path; entity-group replan estimates against the origin DB
- `3d6bf1cb` order a ground-rooted recursive rule as a selective generator (was a full attribute scan; e.g. `indirectly-replaced-by` over a past as-of version 1.2s → 10ms) + regression test

**Native-image release blocker** (CI red on every push since the bitemporal valid-time commit):
- `03c46f87` `next-tx-instant` called `.getTime` on two un-hinted `java.util.Date`s → reflective calls that crash the native binary at the first `transact` (`No matching field found: getTime`). Routed through a hinted cross-platform `tools/date->epoch-ms`.

**Perf**
- `95128230` memoize Date→tx-id resolution for date-based as-of/since (`temporal-info` was re-resolving per op; 233ms → 9ms; numeric-tx as-of unchanged)

`58e889fb` is a cljfmt-only reformat of `execute.cljc`.

Validated: `clj-pss` + `clj-hht` both 623 tests / 0 failures; clean `*warn-on-reflection*` over the touched namespaces.

PR #853, opened 2026-07-04
Bound the query result cache by result size, not just snapshot count

BuddhiLW:
## Problem

`datahike.query/query-result-cache` is an LRU keyed by DB snapshot (`[hash max-tx max-eid]`) and capped by `*query-cache-size*` (default 64 snapshots). Each snapshot bucket, however, holds a full map of cached query results (`{cache-key -> {:result r :attrs deps}}`) whose combined size is unbounded. A workload that keeps a handful of large result sets alive therefore pins arbitrary heap **even though the snapshot-count cap never triggers**.

On a long-running coordinator this cache retained ~1.3 GB of decoded result tuples — 48% of live heap in a heap-dump dominator-tree analysis — while the 64-snapshot cap was never hit.

## Fix

Add `datahike.lru/weighted-lru`: an LRU bounded by **both** an entry-count limit and a cumulative **weight budget**, where a caller-supplied `weigh` fn maps each stored value to its cost. It exposes the same `ILookup`/`Associative` surface as the existing `LRU`, so `query-result-cache`'s `get`/`assoc` sites are unchanged.

`datahike.query` wires the cache to it with a per-snapshot-bucket weigh fn (total cached result tuples) and a new dynamic var `*query-cache-weight-limit*`:

- default **1,000,000** result tuples
- env `DATAHIKE_QUERY_CACHE_WEIGHT_LIMIT`
- `set-query-cache-weight-limit!` at runtime
- `0` disables the weight budget (exact prior behavior)

Eviction removes least-recently-used snapshots until the total is within budget; the most-recent snapshot is always retained, even if it alone exceeds the budget.

## Tests

- `datahike.test.lru-weighted-test` (plain `clojure.test`): count/weight eviction, always-keep-newest, LRU-touch protection, the multi-evict shrink loop, disabled-budget back-compat, and the query-cache bound end-to-end (30 snapshot buckets → 3, retained tuples ≤ budget).
- `datahike.test.query-cache-trifecta-test`: a golden + property + mutation trifecta over `weighted-lru` — 200 generated inputs for the invariants (entry-count cap, weight budget, kept-keys ⊆ put-keys), plus mutants encoding the original count-only bug / never-evict / evict-newest, each caught by the suite. This adds a test-only dependency `io.github.hive-agi/hive-test` (public, git coordinate) to the `:test` alias.

## Compatibility

`weight-limit 0` restores the exact previous behavior (count-only LRU). The 1M-tuple default bounds heap while preserving cache hit rates for typical workloads. CHANGELOG updated under 0.8.


whilo:
Thanks a lot for this, @BuddhiLW — the heap-dump analysis was spot on and `weighted-lru` is a clean, well-shaped fix. It's now landed on `main` via #859, with you credited as author of the fix.

Two adjustments went in on the way:

1. **Rebased onto current `main`.** This branch predated the recent planner / graph / anomaly merges (~13 commits), so the core change (`lru.cljc` + `query.cljc` + CHANGELOG + the plain tests) was cherry-picked forward to keep the diff to just the cache fix.

2. **Test layer swapped.** The `weighted-lru` fix and its `clojure.test` coverage we kept as-is. We dropped the `hive-test` trifecta + golden file + git-coordinate dependency, though — we'd rather not pin CI to an external personal-org SHA, and in this case the golden facet largely duplicated the plain assertions while the mutation facet swapped the *test helper* rather than the production code. In its place is a `test.check` property test (test.check is already a `:test` dep) built around a reference-model oracle that independently re-derives the surviving key set / values / total weight and compares.

That property test earned its keep immediately: it surfaced a latent bug in `bucket-weight` that the original suite missed. `(count (:result entry))` throws on scalar-find queries (`[:find ?x .]`), whose cached result is a bare value rather than a collection — so with the default weight budget enabled, every scalar query broke (it showed up across the persistent-set / hitchhiker-tree / node-cljs suites). #859 fixes it by computing a robust per-result weight **once** at cache-put time and storing it on the entry, so the weigh path never counts a raw result again (scalars weigh 1, collections weigh their tuple count). The weighted-lru tests also now run under the cljs node suite, since the `WeightedLRU` deftype has its own cljs implementation.

Closing this one as superseded by #859 — thanks again, great catch and a genuinely valuable fix. 🙏

BuddhiLW:
Very good, my friend. Next time, I will attend to not inject my `hive-` deps!

Glad it helped!

These benefit me too, and I went through some great lengths to diagnose and improve performance in hive-mcp IAddons that heavily uses datahike and datalevin and a bunch of other things... And indeed, I cut like 50% of RAM usage, after all changes in all projects underneath.

Fertile grounds to battle test datahike! :)

PR #862, opened 2026-07-07
feat(query): let secondary indices own their external-engine query-spec

whilo:
## Problem

The external-engine executor (`execute-external-engine`, `query/execute.cljc`) hardcodes the query-spec passed to a secondary index's `-search`/`-slice-ordered` as `{:query <arg0> :field <arg1>}` in `:filter` and `:retrieval` mode — a text-search shape — even though the `ISecondaryIndex` protocol states query-spec is **index-type-specific (opaque)**. The in-code comment even says *"call the resolved function with args to get query-spec"* but the code doesn't.

This limits any non-text secondary index (e.g. a term / discrimination-tree index over structured values) to two positional args and a misnamed `:query`/`:field` key — the one thing standing between *"works with an awkward shape"* and *"first-class external term index"*.

## Change

An engine may declare an optional `:query-spec-fn` in its `:datahike/external-engine` metadata to build an arbitrary, arity-free spec from the (index-ident-stripped) args:

```clojure
(defn ^{:datahike/external-engine
        {:index-key 0
         :binding-columns [:entity-id]
         :query-spec-fn (fn [args] {:pattern (nth args 1) :depth (nth args 2)})}}
  dt-match [idx pattern depth] ...)
```

Absent it, the executor falls back to the legacy `{:query :field}` shape, so **existing engines (incl. stratum) are unaffected**.

- extract `external-query-spec` (honors `:query-spec-fn`, else the default)
- use it in both `:filter` and `:retrieval` execution modes
- unit test: default backward-compat shape + custom >2-arity opaque spec

## Motivation

Enables an external library to register a first-class term/structural secondary index queryable from datalog **without patching the executor**. (Context: building a CIC discrimination-tree index over declaration types in an external prover, fed incrementally via the existing `ISecondaryIndex` `-transact` signed-delta stream.)

## Testing

New unit test `external-engine-query-spec-test` (2 tests / 6 assertions). No regression: `query-ir` + `stratum-bridge` + `secondary-*` suites green (**43 tests / 141 assertions / 0 failures**).

### DIFF
--- a/src-secondary/datahike/index/secondary/proximum.clj
+++ b/src-secondary/datahike/index/secondary/proximum.clj
@@ -173,3 +173,43 @@
     (assoc key-map
            :branch (name new-branch)
            :commit-id new-commit-id)))
+
+;; ---------------------------------------------------------------------------
+;; KNN as a Datalog clause (external-engine)
+;;
+;; With the query-spec-fn mechanism, a Proximum vector search is a first-class
+;; :where clause the planner recognizes — no manual EntityBitSet plumbing:
+;;
+;;   ;; retrieval — bind entity + cosine distance, then join like any relation
+;;   (d/q '[:find ?name ?distance
+;;          :in $ ?qvec
+;;          :where
+;;          [(datahike.index.secondary.proximum/knn :idx/embeddings ?qvec 10)
+;;           [[?e ?distance]]]
+;;          [?e :doc/name ?name]]
+;;        db query-float-array)
+;;
+;;   ;; filter — entities only (1 binding var)
+;;   [(datahike.index.secondary.proximum/knn :idx/embeddings ?qvec 10) [?e ...]]
+;;
+;; Planner-only: the base (relational) engine has no external-engine mechanism.
+(defn knn
+  "Vector k-nearest-neighbour search over a Proximum secondary index, callable
+   from a Datalog :where clause. Args: <idx-ident> <query-vector (float[])> <k>.
+   Binds [[?e ?distance]] (retrieval) or [?e ...] (filter). See the ns comment."
+  {:datahike/external-engine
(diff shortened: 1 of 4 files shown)


whilo:
Rebased onto `main` (now includes #896 `:db.type/float-array`), fixed the `persistent-set-test-base-engine` failure, and added the first real consumer of the `query-spec-fn` mechanism — **Proximum KNN as a Datalog clause**.

**Fixes**
- `persistent-set-test-base-engine`: the external-engine test errored under the base engine (`Cannot bind value true to [[?e]]`) because external-engine clauses are a *planner* feature — the base engine has no mechanism to recognize the metadata. Added a `use-fixtures` forcing the planner on for the d/q-driven tests.
- Retrieval mode mapped result columns with `(keyword (name var))`, keeping the leading `?` (`?distance` → `:?distance`), so every non-entity column came back `nil`. Now maps positionally by the engine's declared `:binding-columns` (e.g. `[:entity-id :distance]`), with a `?`-stripped-name fallback.

**New — Proximum `knn` clause** (demonstrates the query-spec-fn payoff):
```clojure
;; retrieval: entity + cosine distance, then join like any relation
[(datahike.index.secondary.proximum/knn :idx/emb ?qvec 10) [[?e ?distance]]]
;; filter: entities only, composes with arbitrary Datalog constraints
[(datahike.index.secondary.proximum/knn :idx/emb ?qvec 10) [?e ...]]
```
`knn` carries `:datahike/external-engine` metadata whose `:query-spec-fn` builds Proximum's `{:vector … :k …}` shape instead of the full-text `{:query :field}` default. Tests cover both modes; green under `DATAHIKE_QUERY_PLANNER=true` and `false` (15 tests / 71 assertions each).

PR #864, opened 2026-07-07
feat(rdf): RDF / RDF-star via interned literals (federated named graphs)

whilo:
## Motivation

RDF / semantic-web modelling on datahike reduces to **one** genuinely new primitive: a native literal value type. RDF's other node kinds already map onto datahike — an **IRI** is a `:db/unique` entity (a ref, addressed by lookup-ref, the same primitive as [cross-database references](https://github.com/replikativ/datahike/blob/main/doc/cross-db-references.md)), a **blank node** is an eid, and a **quoted triple** is a reified statement — so the only shape datahike lacks is a **literal** `(lexical, datatype, lang)`.

Rather than special-case RDF in core, this PR adds a **general seam for custom value types** and makes RDF its first client. This is useful on its own: users can add a `:db/valueType` (geo, money, tensors, …) without hacking datahike.

## What this adds

**`datahike.value-types`** — a registry seam. Register a `:db/valueType` with its predicate, `Comparable` ordering, and fressian/transit/edn codecs:

```clojure
(vt/register! :db.type/mytype
  {:pred my-pred? :describe "…"
   :fressian-read {…} :fressian-write {…}   ; durable store
   :transit-read {…}  :transit-write {…}     ; remote/HTTP (optional)
   :edn-tag 'my/tag   :edn-reader my-reader}) ; pr/read (optional)
```

- **Comparator is unchanged.** `datahike.datom/compare-value` ends in `(compare v1 v2)`, and `compare` delegates to `Comparable` — so a `Comparable` value sorts in AVET with **zero change to the comparator and zero cost to the scalar path**. The value's `compareTo` *is* the on-disk order.
- **Fail-loud cross-peer contract.** The `:db/valueType` keyword persists in the stored schema and travels to every peer for free; the impl does not. `conn-from-db` calls `assert-registered!`, so opening a store whose schema names an unregistered custom type fails **at connect** with a clear error, not deep in a later read.

**`datahike.experimental.rdf`** — the first client:
- **`:db.type/literal`** — a native RDF literal, sorted in exact SPARQL literal order (numerics compare numerically via `bigdec`; else lexical → datatype → lang). RDF 1.1 term equality. `literal` / `literal-attr` helpers.
- **RDF-star statement reification** via composite tuples: a triple's identity is the auto-derived, unique `[s p o]` tuple, so a statement is addressable by lookup-ref, annotatable, and idempotent on re-assertion (RDF set semantics). Corresponds to RDF 1.2's `rdf:reifies` reifier.
- `rdf-term-compare` — the full SPARQL ORDER BY order over mixed terms, for ordering query results.

## Core touches (all additive)

- `value_types.cljc` (new) — the registry.
- `schema.cljc` — `:db.type/value` spec consults the registry (builtin ∪ registered); `describe-type` keeps the helpful builtin list.
- `index/persistent_set.cljc` — merge registered fressian handlers into the store serializer (the durability path).
- `remote.cljc` / `http/client.clj` — merge registered transit/edn handlers.
- `connector.cljc` — connect-time registration check.

No comparator change, no rschema change, no change to existing stores.

## Verification

Validated end-to-end on a file backend: durable **write → release → reconnect** reconstructs `RDFLiteral` values from disk and preserves datatype-aware order (`9 < 42`, not `"42" < "9"`); AVET reverse-lookup by a full literal survives reload; transit + EDN round-trip; connect-time check passes for a registered type and throws for an unregistered one. Core suites (core/schema/tuples) pass (24 tests / 137 assertions, 0 fail / 0 error). `clj -M:ffix` clean.

See [`doc/rdf.md`](https://github.com/replikativ/datahike/blob/feat/rdf-custom-value-types/doc/rdf.md) and `datahike.experimental.rdf/demo` for a runnable showcase.

## Scope / limits

- **Experimental** — shipped to gather feedback on the complete package; the literal wire format may change.
- **CLJS**: the literal type + comparison + equality are cross-platform; the durable fressian/transit codecs are currently CLJ-only (cljs persistence is a follow-up). In-memory cljs use works.
- OWL-DL reasoning and a SPARQL parser are out of scope; RDFS/OWL-RL-lite entailment is expressible via datahike's recursive rules. Named graphs map to store-per-graph via the reference layer.


whilo:
## Reconciling with #881 — rework to interned-literal entities (option C)

#881 removed the `datahike.value-types` registry (open value-type extension), so this PR can't merge as-is. Evaluating the reconciliation against Virtuoso and the SPARQL 1.1 spec turned up that the native `:db.type/literal` isn't just excess core weight — **it's a correctness bug**, so the fix is to drop it rather than rebase it.

**The bug.** `rdfliteral-compare` returns `(compare nka nkb)` for two numeric literals with **no lexical/datatype tiebreak**. `numeric-key` of `"1"^^xsd:integer`, `"01"^^xsd:integer` and `"1"^^xsd:long` are all `(bigdec "1")`, so all three `compareTo` as **0**. The AVET index is a sorted set keyed by `compareTo`, so it collapses three distinct RDF *terms* into one slot — while `equals`/`hashCode` key on lexical form and disagree. Result: a cardinality-many attribute loses a datom, and `COUNT`/`DISTINCT`/point-lookups return wrong SPARQL results. Root cause is structural: the value type **fuses ordering and identity into one `Comparable`, then sacrifices identity for numeric order** — but RDF's index must key on *term* (lexical ∧ datatype ∧ lang), never merging distinct terms.

**The rework (option C):** a literal becomes an **interned entity** keyed by its canonical term hash — which is exactly Virtuoso's `RO_ID` model, gives term dedup for free, and makes the collapse structurally impossible. Numeric range becomes `d/index-range` on a `:literal/num` (`:db.type/bigdec`) shadow attribute — a homogeneously-typed key datahike already seeks, with no RDF code in the core comparator. `rdf-term-compare` stays for query-side `ORDER BY`.

**The irreversible part** is the canonical term-id function (lang→lowercase, plain→`xsd:string`, lexical **verbatim**) — it becomes on-disk identity, so it needs tests pinning all four rules before any data is written.

**Named graphs:** decided to ship **federation** (one store = one dataset; named graphs = federation via the reference layer) as a documented 1.0 scope limit, *not* a quad. A 4th indexed datom position is the alternative and was consciously deferred (it's a storage-format change, revisitable post-1.0).

Full design, the interned-entity schema, and the faithfulness fixes (reified-triple object typing, `triple`/`annotate` asserting the ground triple, bnode scoping, term-kind recoverability) are written up for whoever picks this up. This does **not** need a value-type seam — RDF literals were the one arguable case for it, and interning serves them better *and* correctly.

PR #866, opened 2026-07-08
fix(native): reflection warnings during native builds

harismh:
#### SUMMARY

Fixes the remaining reflection warnings when building native images. As of writing builds finish without any reflection warnings. 

##### Bugfix
- [x] Formatting checked


src/datahike/audit.cljc
```diff
@@ -148,9 +148,9 @@
                       _       (when-not head
                                 (throw (ex-info "verify-chain: no head cid on db"
                                                 {:type :audit/no-head})))
-                      commits (volatile! [])
-                      missing (volatile! [])
-                      visited (volatile! #{})]
+                      ^clojure.lang.Volatile commits (volatile! [])
+                      ^clojure.lang.Volatile missing (volatile! [])
+                      ^clojure.lang.Volatile visited (volatile! #{})]
```
harismh: I believe unknown type hints are benign in CLJS, but I could swap to a reader conditional if wanted.

PR #868, opened 2026-07-10
feat(gc): background concurrent mark-and-sweep + sweep grace window (experimental)

whilo:
## What

Two additions to `datahike.gc`, extracted from this week's GC stress campaign (context in #867's discussion):

1. **`:grace-ms` on `gc-storage!`** — the sweep cutoff becomes `now − grace` instead of GC-start time: objects written within the grace window are never swept, regardless of reachability. This closes the concurrent-writer race (the mark walk reads each branch head once; a commit flipping a head after that read may reference objects the walk never saw) and doubles as the reader-horizon knob: datahike's reader model is **open** (no registration), so grace is sized to the longest expected reader/query horizon. Optional reader leases for long-running distributed readers can layer on later without changing this contract. Default 0 = exactly today's behavior.

2. **`start-background-gc!`** — runs `gc-storage!` on an interval with the grace guard and optional `:history-window-ms` ranged commit pruning; returns a stop fn.

## Positioning: two GC tiers that compose

- **Online freed-tracking GC** (existing) — the *incremental* tier: O(garbage) work, freelist recycling keeps object/file counts flat. Right for single-branch bulk imports (the bett/dbpedia use case). With persistent-sorted-set#16 its tracking is exact for serial writers.
- **This collector** — the *completeness* tier: full reachability, so it supports **multiple branches** (online GC refuses them), needs no mutation-time tracking (reclaims what freed-tracking structurally cannot see — measured ~94% of superseded objects under a pipelining writer), and prunes commit-graph records (which online GC never touches). Cost: an O(reachable) walk per cycle — hence background, low frequency.

## Validation

Compressed into the regression tests from longer stress runs:
- **133,881 pipelined txs at ~3k tx/s with 3 concurrent collection cycles** → store contained to 193 files (fewer than it started with), model exact warm and after cold reopen.
- **Two divergent live branches under concurrent writes** → 11k objects swept, both branches byte-exact warm+cold, history intact.
- Existing gc/online-gc suites: 42 tests / 0 failures (the 2-arity behavior is unchanged).

*Experimental.* Independent of #867 (no file overlap in conflict).


whilo:
Added the stress harnesses used to validate this PR: `benchmark.soak` (four concurrent model-checked scenarios — oltp+background-GC, bulk-import+online-GC, branching, crypto audit — overnight-capable via `SOAK_MINUTES`, non-zero exit on any invariant failure) and `benchmark.gc-stress` (write-load × GC-strategy grid). The soak's first 2-minute smoke immediately found persistent-sorted-set#17 (a pre-0.4.122 baseline-store crash under heap pressure); with all current fixes it runs all-green.

PR #889, opened 2026-07-19
fix(query): recursive rules with a ground argument silently truncate at 1000 results

whilo:
## The bug

A recursive rule called with a **ground argument** over a ref edge — the canonical shape:

```clojure
(d/q '[:find [?a ...] :in $ % ?h :where (anc ?h ?a)]
     db
     '[[(anc ?c ?a) [?c :parent ?a]]
       [(anc ?c ?a) [?c :parent ?m] (anc ?m ?a)]]
     head-eid)
```

**silently returns exactly 999 rows** whenever the true closure is deeper than 1000. No error, no warning — plausible-but-wrong results. (Found running Datalog over a 3,586-commit git history: ancestors of HEAD came back as 999.)

Root cause: the magic-set demand restriction was a runtime bolt-on that rebuilt the demand relation from its HashSet **every fixpoint iteration** (`inject-magic-relation` — O(|demand|) per round, quadratic overall). To escape that cost it carried an **explosion guard**: past 1000 demanded entities it abandoned the restriction. On the abandonment path the single-ref-edge base (`base-scan-attr`) was excluded from the full-base backfill, so base facts for entities demanded past the threshold were never produced — the frontier starves and the fixpoint terminates early at threshold−1.

Note the benchmark suite's recursive-rule queries are all **unbound** `(follows ?e ?e2)`; the ground-argument quadrant was untested, which is how this shipped.

## The fix

Make the demand set a first-class, incrementally-maintained relation — the shape a static magic-set rewrite produces (cf. Soufflé, where the demand predicate is an ordinary relation in the same semi-naïve fixpoint) — and keep the restriction **always on**:

- **Demand index + delta batch**: each value is admitted once (O(1)); first-time values form a per-iteration batch, the only thing the next round's base scan consumes. Base work is O(|demand|) *total*, not per round (previously the base scan also re-iterated the entire demand set every round).
- **The guard is an O(1) membership semijoin** (`demand-filter-rel`) on derived tuples, applied *before* dedup so dropped tuples stay re-derivable. Replaces the per-round demand-relation rebuild.
- **Deleted**: `magic-explosion-threshold`, the abandonment fallback and both backfill branches, the per-iteration `inject-magic-relation` call. There is no cost cliff left to escape, hence no guard, hence no truncation mode.

Completeness argument for the guard (in the `demand-filter-rel` docstring): demand spreads along base edges at least as fast as breadth-first derivation, so every shortest-path derivation chain reaching the ground result row has its source demanded by the time the deriving delta arrives; longer-path derivations it drops are redundant for that row.

## Verification

Correctness vs BFS ground truth (all exact):

| shape | sizes |
|---|---|
| chains, card-one + card-many | 500 / 1001 / 1500 / 5000 |
| binary tree leaf ancestry | depth 13 |
| merge DAGs (diamonds) | 1200 / 3600 nodes |
| sibling-noise (undemanded entities edging into the demanded region) | 50 / 500 / 1500 |
| uneven diamond (short + long path rejoining) | up to 1200-long rejoin |

- `datahike.test.query-rules-test` + `datahike.test.query-planner-test`: **0 failures**.
- New regression test `test-deep-bound-recursive-rule` (deep bound chain, interior-bound, sibling-noise, uneven diamond) — **fails on current main** at exactly the truncation and guarding cases, passes with the fix.
- cljfmt clean.

Performance (uncached, fresh stores; vs the abandonment path, which additionally computed the closure of the *whole* database):

- chain 500: 341 → **133 ms**; chain 5000: 92.0 → **40.1 s** (the O(closure) relational bound is intrinsic to materializing TC; the win here is correctness + staying on the reachable subgraph)
- **ancestors of one 100-chain head in a 200,000-entity forest: 47 ms** — the fixpoint never leaves the reachable component; the fallback would have computed every chain's closure.
- 20×2500-node forest, one 2499-deep ancestry (past the old threshold): 11.2 s touching one chain instead of twenty.

## Follow-ups (not this PR)

- Add ground-argument recursive queries to the cross-database benchmark suite (the untested quadrant).
- Single-source reachability is intrinsically cheaper than materializing the TC relation — dispatching this query shape to `datahike.experimental.graph/transitive-closure` (O(V+E)) once the graph API is stable would take the 47 ms case to sub-ms and the deep-chain case from quadratic to linear.

### DIFF
--- a/src/datahike/query/execute.cljc
+++ b/src/datahike/query/execute.cljc
@@ -3394,13 +3394,15 @@
      :cljs nil))
 
 (defn- magic-base-scan
-  "Directly scan edges for entities in the demand set using EAVT point lookups.
-   Only scans entities not yet in the scanned set (if provided).
-   Much faster than re-executing the full base branch plan for large graphs.
+  "Directly scan edges for the newly demanded entities using EAVT point lookups
+   — the demand relation's delta step. `batch` holds only first-time demand
+   values (admission into the demand index is the dedup), so each entity is
+   scanned exactly once and total base work stays O(|demand|) across the whole
+   fixpoint. Much faster than re-executing the full base branch plan.
    Only works for simple binary base patterns like [?x :attr ?y]."
-  [db head-vars attr demand-set scanned-set]
+  [db head-vars attr ^java.util.ArrayList batch]
   #?(:clj
-     (when attr
+     (when (and attr (pos? (.size batch)))
        ;; Resolve the EAVT index through any temporal wrapper: AsOfDB/SinceDB/
        ;; HistoricalDB carry no :eavt of their own (it lives on the origin-db),
        ;; and the slice must merge current+temporal history and post-process by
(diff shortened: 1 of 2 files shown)


whilo:
Force-pushed a revision after the planner-regression suite (thanks @whilo) flagged a real slowdown in the first cut.

**What the regression was:** the first version replaced the old demand *push-down* (inject the demand relation so the recursive scan only touches demanded `?ground`) with a *post-filter* that ran after `execute-branch-plans` had already expanded every reverse-neighbor. For a *selective* ground query — `(reaches ?start ?b)` rooted near the end of a chain, ~49 reached — that meant computing the whole chain's closure and then throwing almost all of it away. `recursive-ground` came out 3.3x **slower** than the planner-off baseline; `asof-recursive` 2.0x.

**The fix:** keep the always-on / incremental design, but restore the push-down — inject the accumulated demand each round to prune `?ground` *before* expansion. The key is that the demand relation is grown by **append** (a one-column tuple per admitted value, mirroring the demand index), so wrapping it each round is O(1) and it is never rebuilt from the index — which is what made the old push-down O(|demand|) per round and forced the explosion guard in the first place. The post-filter is gone.

Planner-regression now, same store (4000 nodes, 300-tx history):

| shape | ratio (compiled/base) | was |
|---|---|---|
| recursive-ground | **0.30x** | 3.32x SLOW |
| asof-recursive | **0.71x** | 2.04x SLOW |

All shapes OK, results agree. Rule + planner kaocha suites: 93 tests / 399 assertions / 0 failures (incl. the new deep-bound regression test). Correctness re-verified against BFS ground truth on all the shapes above.

Good catch — the selective-query push-down is exactly the property the benchmark exists to protect.

PR #893, opened 2026-07-20
fix(secondary): Proximum bridge coerces :db.type/tuple vectors to float[]

whilo:
## Problem

The **declarative** path for a Proximum vector index — exactly what `doc/secondary-indices.md` documents — could not index anything.

A `:db.secondary/attrs` embedding attribute is declared `:db.type/tuple`, whose values datahike validates as `vector?`. But the Proximum bridge's `-transact` only inserted values that were already a `float[]`:

```clojure
(if (instance? (Class/forName "[F") val)
  (prox/insert prox-idx val eid)
  (log/warn :datahike/non-float-array-vector ...))   ; silently skipped
```

The two contracts have **no overlap**:

- a Clojure vector passes write-time validation → bridge skips it (warn, 0 indexed)
- a `float[]` is what the bridge wants → fails `:db.type/tuple` validation before it reaches the index

So a schema-declared Proximum index silently indexed nothing. The bug never surfaced because the existing tests all use the manual `sec/create-index` + `db/empty-db` path with raw `float[]`, bypassing validation.

Reproduced verbatim from the doc (only the value shape changed to a vector, since a `float[]` cannot validate):

```clojure
(d/transact conn [{:person/embedding [1.0 0.0 0.0 0.0]} ...])
;; before: 0 vectors indexed
```

## Fix

Coerce at the bridge boundary. `-transact` now accepts a `float[]` verbatim (unchanged for the manual path) and coerces a vector/seq of numbers to `float[]`. The primary index keeps the immutable vector (hashes by value); Proximum gets the primitive array it needs.

- `doc/secondary-indices.md`: the example transacted `(float-array …)`, which cannot validate against `:db.type/tuple` — corrected to plain vectors, with a note on why.
- Added a regression test exercising the full declarative path (create-database → connect → schema-transact index → transact vector embeddings). It **fails before this change** (0 indexed → cardinality assertion fails) and passes after.

## Validation

- Full `secondary-integration-test`: **12 tests, 59 assertions, 0 failures** (was 11/58; +1 regression test).
- Confirmed the new test catches the bug: reverting only the source change → **1 failure**.
- `float[]` passthrough, int/double-vector coercion, and non-numeric rejection all verified.

Found while building a semantic-code-search demo over a geschichte repository (Clojure's history), where the declarative Proximum index is the natural integration point.

### DIFF
--- a/doc/secondary-indices.md
+++ b/doc/secondary-indices.md
@@ -109,10 +109,13 @@ Proximum provides HNSW-based approximate nearest neighbor search. Requires Java 
                                                         :id (random-uuid)}}}])
 (Thread/sleep 1000)
 
-;; Add vector data
-(d/transact conn [{:person/embedding (float-array [1.0 0.0 0.0 0.0])}
-                  {:person/embedding (float-array [0.0 1.0 0.0 0.0])}
-                  {:person/embedding (float-array [0.9 0.1 0.0 0.0])}])
+;; Add vector data. A `:db.type/tuple` value is a plain Clojure vector — that
+;; is what datahike validates and stores in the primary index; the Proximum
+;; bridge coerces it to a float[] for the HNSW graph. (Query vectors passed to
+;; `-search` below are float[], the shape Proximum's search API takes.)
+(d/transact conn [{:person/embedding [1.0 0.0 0.0 0.0]}
+                  {:person/embedding [0.0 1.0 0.0 0.0]}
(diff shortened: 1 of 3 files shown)


whilo:
Superseded by #896, which adds first-class `:db.type/float-array` / `:db.type/double-array` value types. That is the right fix: a float array validates and is indexed by the existing Proximum bridge directly — no tuple representation and no coercion. Closing in favor of #896.

PR #894, opened 2026-07-20
fix(transact): omit no-op re-assertions from :tx-data (#457)

### LINKED ISSUE #457: [Bug](?): Transact result :tx-data no longer shows only changed datoms

0.4.1473

11.0.12

Linux

(def config
    {:schema-flexibility :read
     :store {:backend :file
             :path (.getAbsolutePath file)}})

### Describe the bug

I have a previous project using datahike version 0.2.1. In this project when I `transact` the exact same transaction twice in a row, `:tx-data` is empty except for the transaction entity the second time around. i.e. It only shows me what has changed as a result of the transaction. This is super useful, as I can use `datahike` to produce diffs for me of previous vs new DB state!

Now I'm starting a new project using the latest version and this behavior is gone. It shows me all datums every time even when they are noops.

Is this intentional? I really liked the old behavior. :)

Awesome project, BTW!

Example:

0.2.1
```
shazoom.db=> (clojure.pprint/pprint (d/transact (conn) [[:db/add 2 :foo :bar]]))
{:db-before
 #datahike/DB {:schema ...}
 :db-after
 #datahike/DB {:schema ...},
 :tx-data
 [#datahike/Datom [536871450
                   :db/txInstant
                   #inst "2021-12-16T00:46:05.195-00:00"
                   536871450]
  #datahike/Datom [2 :foo :bar 536871450]],
 :tempids #:db{:current-tx 536871450},
 :tx-meta nil}
nil
shazoom.db=> (clojure.pprint/pprint (d/transact (conn) [[:db/add 2 :foo :bar]]))
{:db-before
 #datahike/DB {:schema ...},
 :tx-data
 [#datahike/Datom [536871451
                   :db/txInstant
                   #inst "2021-12-16T00:46:07.024-00:00"
                   536871451]],
 :tempids #:db{:current-tx 536871451},
 :tx-meta nil}
nil
```

0.4.1473
```
user=> (clojure.pprint/pprint (db/transact [[:db/add 1 :foo :bar]]))
{:db-before #datahike/DB {:max-tx 536871012 :max-eid 678},
 :db-after #datahike/DB {:max-tx 536871013 :max-eid 678},
 :tx-data
 [#datahike/Datom [536871013
                   :db/txInstant
                   #inst "2021-12-16T00:38:34.750-00:00"
                   536871013]
  #datahike/Datom [1 :foo :bar 536871013]],
 :tempids #:db{:current-tx 536871013},
 :tx-meta nil}
nil
user=> (clojure.pprint/pprint (db/transact [[:db/add 1 :foo :bar]]))
{:db-before #datahike/DB {:max-tx 536871013 :max-eid 678},
 :db-after #datahike/DB {:max-tx 536871014 :max-eid 678},
 :tx-data
 [#datahike/Datom [536871014
                   :db/txInstant
                   #inst "2021-12-16T00:38:38.672-00:00"
                   536871014]
  #datahike/Datom [1 :foo :bar 536871014]],
 :tempids #:db{:current-tx 536871014},
 :tx-meta nil}
nil
```

Only changed datums shown in :tx-data

Transact the same transaction twice and observe datums are same both times

jgertm:
Fixes #457.

## Problem

Re-asserting `[:db/add e a v]` for a value the entity already holds (cardinality-one) or already carries among its values (cardinality-many) still emitted that datom into the transaction report's `:tx-data`. So two identical back-to-back transactions no longer produced an empty data diff — the property (present in ≤ 0.2.1, like DataScript) that lets `:tx-data` be read as *what actually changed*.

The skip-when-unchanged guard was dropped during the mid-2020 upsert refactor (`e9d55972`, "Implements a big part of the solution towards upsert"), which replaced the guarded cardinality-one branch with an unconditional emit and never restored the no-op semantics.

## Fix

`transact-report` now omits the datom from `:tx-data` when the exact `[e a v]` is already live, comparing through the index's own datom comparator (`di/-slice`) — so it is content-correct for `:db.type/bytes` and tuple values, where Clojure `=` would compare byte arrays by identity.

**Only the `:tx-data` entry is suppressed.** The write itself still runs, so:
- secondary indices (incl. bitemporal SCD2) still receive the assertion,
- history and composite-tuple materialization are unaffected,
- a genuine change still asserts, cardinality-one still replaces the prior value,
- retracting an absent value is unchanged.

## Tests

`test-reassert-datom-is-noop` (9 assertions): cardinality-one/many no-op and change, re-asserting one of many cardinality-many values (leaving the others intact), tuple-valued attributes, and `:db.type/bytes` (content comparison).

Full `:clj-pss` and `:clj-hht` suites: 794 tests / 6751 assertions / 0 failures on both index backends.

## Not included

An orthogonal, pre-existing bug surfaced during review: incremental `:hash` on a genuine cardinality-one update adds the new datom's hash without subtracting the replaced one, so it diverges from a fresh rebuild (breaking the hashCode/equals contract). Left out to keep this fix minimal — worth a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


whilo:
@jgertm Danke! I have merged this with #946.

Issue #901, opened 2026-07-25
Query planner NPE in post-filter-not-joins on a not-join with :in-bound vars (0.8.1747; distinct from #897)


markaddleman:
## Summary

On **0.8.1747**, a query containing a `not-join` whose variables are supplied through `:in` throws a `NullPointerException` in the **query planner** path (`datahike.query.execute/post-filter-not-joins`). Disabling the planner (`DATAHIKE_QUERY_PLANNER=false`) returns correct results.

This is **distinct from #897** (the recursive-rule `rel-dedup-into!` NPE, fixed by #899 — thank you, confirmed fixed on 0.8.1747). Same planner-path family, different site: this one is a `not-join`, no recursion involved.

## Minimal repro

```clojure
(require '[datahike.api :as d] '[datahike.query :as dq])

(let [cfg {:store {:backend :memory :id (java.util.UUID/randomUUID)}
           :schema-flexibility :read :keep-history? false}]
  (d/create-database cfg)
  (let [conn (d/connect cfg)]
    (d/transact conn [{:db/id -1 :triple/s "s1" :triple/p "isa" :triple/o "C"}
                      {:db/id -2 :triple/s "s1" :triple/p "P"   :triple/o "v"}
                      {:db/id -3 :triple/s "s2" :triple/p "isa" :triple/o "C"}])
    (let [db (d/db conn)
          q  '[:find ?s
               :in $ ?C ?P
               :where
               [?e :triple/s ?s] [?e :triple/p "isa"] [?e :triple/o ?C]
               (not-join [?s ?P]
                         [?e2 :triple/s ?s] [?e2 :triple/p ?P])]]
      (println "planner default:" (try (d/q q db "C" "P")
                                        (catch Throwable t [:threw (.getMessage t)])))
      (binding [dq/*disable-planner* true]
        (println "planner off:    " (d/q q db "C" "P"))))))
```

Output on 0.8.1747:

```
planner default: [:threw Cannot invoke "java.lang.Character.charValue()" because "x" is null]
planner off:     #{[s2]}
```

`#{[s2]}` is correct: `s2` `isa C` but has no `P` predicate; `s1` has `P`, so it is excluded by the `not-join`.

## Stacktrace (planner on)

```
java.lang.NullPointerException: Cannot invoke "java.lang.Character.charValue()" because "x" is null
  at clojure.lang.RT.intCast (RT.java:1241)
[...]

PR #906, opened 2026-07-26
test(query): generate the :in binding shapes the differential test never did

whilo:
The generative differential test bound only `[?e ...]` over entity ids, `%` and `$2`. Every divergence in the #901 family needed something else, because a **scalar `:in` binding** is what makes the two engines see different clause text: the planner folds its value into the clauses before planning. That blind spot is why all of them survived the existing suite.

Adds, and crosses with the existing axes: scalar `:in` (including one bound to a var no clause mentions — the shape that proves a post-fold decision cannot be cached), tuple and relation bindings, lookup-ref values, rules called with a ground argument on either side, `:with`, and `:order-by`.

Two dimensions target the anti-join's blind spots specifically:

- **a disjunction nested inside a negation**, whose satisfied-but-column-less branch has to survive being handed a context with no relations;
- **`:bare-negation`**, which drops every positive pattern so the only var comes from `:in` and the negation really is evaluated against no relations at all.

Both were unreachable before: every generated query emitted `[?e :name ?n]`, which always leaves a relation behind. Reverting the corresponding fix in the parent PR makes this suite fail immediately, so the dimensions have teeth.

Deliberately **not** generated: `:limit` / `:offset`. With ties in the sort key the row set at a cut-off is genuinely ambiguous, so a disagreement there would not be a bug. `:order-by` is compared as a multiset for the same reason — it still catches wrong or missing rows. `:in` and args are assembled from one ordered list of pairs so they cannot drift apart.

Known remaining gaps, for whoever extends this next: `:attribute-refs? true`, three or more sources, `pull` with a variable pattern, and mutual recursion with ground args. A nested `d/q` also returns a `QueryResult` on the planner path and a set on the base path — equal at top level but not nested inside a tuple — so any nested-`q` dimension would report false divergences until that is reconciled.

3000 cases in CI (~23s); `DATAHIKE_DIFF_CASES=8000` verified locally.

Stacked on #905; fails without all three parents.


whilo:
Folded into #905, which is now based on `main` rather than stacked behind it. The generator test exists only to pin #905's fixes, so it belongs in the same change — and stacking PRs on feature branches is what stranded #904's content when its parent was squash-merged (see #907).

Issue #531, opened 2022-05-05
[Bug]: restore operation fails (with no transaction history)


awb99:
### What version of Datahike are you using?

0.4.1480

### What version of Java are you using?

openjdk 17.0.1 2021-10-19

### What operating system are you using?

linux, guix

### What database EDN configuration are you using?

{:store {:backend :file
                  :path "data/datahike-db"}
          :keep-history? false}

### Describe the bug

I run export-db from a running server.
Then I start on another machine with a blank new datahike database, and add a schema transaction, 
then I run import-db from the same dump that I created earlier.

On the restore there are this errors:
```
Cannot add  #datahike/Datom [36 :db/ident :fulfillment/delivery-date 538003659 true]  because of unique constraint:  #datahike/Datom [39 :db/ident :fulfillment/delivery-date 536870913 true] 

{:error :transact/unique, :attribute :db/ident, :datom #datahike/Datom [36 :db/ident :fulfillment/delivery-date 538003659 true]}

Error during transaction Cannot add #datahike/Datom [36 :db/ident :fulfillment/delivery-date 538003659 true] because of unique constraint: #datahike/Datom [39 :db/ident :fulfillment/delivery-date 536870913 true]
```

This is an extract from the dump:
```
#datahike/Datom [36 :db/cardinality :db.cardinality/one 538003659 true]
#datahike/Datom [36 :db/ident :fulfillment/delivery-date 538003659 true]
#datahike/Datom [36 :db/valueType :db.type/instant 538003659 true]
#datahike/Datom [39 :db/cardinality :db.cardinality/one 538003659 true]
#datahike/Datom [39 :db/ident :fulfillment/courier 538003659 true]
#datahike/Datom [39 :db/valueType :db.type/string 538003659 true]
```

There are a couple hundred lines like this in the beginning. I assume it is the schema specification

Is a restore from a dump expecting that I dont commit my schema first?

Is it possible to do a restore that does not import the schema?

The import works when I delete all the "schema lines" in the dump manually before importing.

### What is the expected behaviour?

I have setup the datahike services so that it creates my schema when there is no currently existing database directory
automatically. So what would be nice is to do a db restore WITHOUT including the schema transactions.
This would make sense because if I do not keep a transaction log, and I have schema changes over time,
then the transactions from the dump would fail.

So what I want to do is make sure that when I import from a dump, I have certainity that my current schema is
satisfied from all imported transactions.

[...]

Issue #508, opened 2022-04-06
[Bug]: 


jsmassa:
### What version of Datahike are you using?

0.4.1489

### What version of Java are you using?

openjdk 11.0.13 2021-10-19

### What operating system are you using?

Ubuntu

### What database EDN configuration are you using?

`{:attribute-refs? true}`

### Describe the bug

The export for reference databases doesn't export system datoms, which then cannot be read in during the import and used to set the system schema. At the moment it works as it is now, but in case the system entities change, we will run into issues. 

### What is the expected behaviour?

The system entities are also exported and the import of reference databases sets the system schema just as it currently sets the user schema from datoms.

### How can the behaviour be reproduced?

Change the order of the system entities between export and import and see what happens.

Issue #377, opened 2021-08-24
adjust import/export for tx log


kordano:

jsmassa:
Connected to #361

PR #909, opened 2026-07-27
feat(migrate): robust, type-exact, verifiable export/import

alekcz:
## Robust, type-exact, verifiable export/import for `datahike.migrate`

Reworks `datahike.migrate` from a flat CBOR EAVT snapshot into a full-history,
type-exact, verifiable dump/restore. New 3-arity `export-db`/`import-db` produce and
consume an **EDN-lines** format (closed reader map — no `read-string`, no eval) and
import through `load-entities` (ids are remapped, so a restore is *semantically*
equal, not id-identical). The legacy 2-arity still **reads** old CBOR dumps.

Full design + rationale: **`doc/import-export-design.md`**. Operator guide:
**`doc/backup.md`**.

### Issues addressed (all verified end-to-end)

| Issue | Fix |
|---|---|
| [#633](https://github.com/replikativ/datahike/issues/633) | `:db.type/double` round-tripped back as `Float` — CBOR routes zero/NaN/±Inf doubles through half-precision (decoded via `Float/intBitsToFloat`), and datahike's schema is a bare `double?`/`float?` predicate with no coercion, so the class flipped. New codec encodes by runtime class with tags (`#datahike/float`, `/bytes`, `/farray`, `/darray`, `/sysref`) and round-trips every builtin value type class-exactly, incl. `float-array`/`double-array`. |
| [#377](https://github.com/replikativ/datahike/issues/377) | Full history — assertions, retractions, tx entities incl. `:db/txInstant`; `history`/`as-of`/`since` match at every `t`. |
| [#262](https://github.com/replikativ/datahike/issues/262) | Schema/refs replay before use (tx-ordered; `:db/txInstant` first within a tx; source order otherwise). |
| [#508](https://github.com/replikativ/datahike/issues/508) / [#531](https://github.com/replikativ/datahike/issues/531) | Attribute-refs dbs **translate** refs to system entities by ident (seed the `:migration` id-map with target system-entity identity) instead of re-inserting system datoms — no `:db/ident` collisions. |
| [#287](https://github.com/replikativ/datahike/issues/287) | `max-tx` is maintained by `load-entities`; the file-scan heuristic is deprecated. |

Also adds: per-chunk **SHA-256** + an order-independent **semantic digest**;
`verify` and `finalize-import!`; chunked-directory and flat formats (manifest
written last as the commit marker); secret allowlist for `:source-config`;
chunk-path validation; owner-only file perms; a **recreate-and-restart** policy for
the (non-resumable) import.

### Grounded against live source (corrects a prior internal spec)

The design was written from an abridged reading of the code; grounding it against
`main` invalidated several assumptions, all corrected here:

- **No `read-string` RCE** — the importer uses CBOR, not `read-string`; there was no
  eval hole to "fix".
- **Retract-first sort is unsafe** — the direct transactor is order-sensitive (a
  card-one add upserts; a retract is an exact `[e a v]` no-op if already replaced),
  so retractions are **not** globally sorted before assertions.
- **No forward-ref core change** — `transact-entities-directly` already allocates a
  target eid on a ref-value miss.
- **Import resume is impossible** — `:migration` is memory-only (`db->stored` omits
  it), so a partial import is recreated-and-restarted, not resumed.

### Tests

`test/datahike/test/migrate_test.clj` — **12 tests, 39 assertions, 0 failures** on
the persistent-set and hitchhiker-tree indices and under spec instrumentation;
cljfmt clean. Covers round-trip of every value type (flat + chunked), the #633
regression, attribute-refs round-trip, schema-ordering, unknown-tag / no-eval,
bad-chunk-path, tamper (sha256) detection, non-empty-target refusal, determinism,
and legacy-CBOR import.

### Known follow-ups (documented in design §14, not overclaimed)

- Export sorts **in memory** (correct, but not yet the 40M-datom bounded-memory
  target — external merge-sort slots behind `export-records` without a format
  change).
- `verify` implements tiers 0–1 (checksums + counts); tiers 2–3 designed, not wired.
- `:on-error :collect` is batch-granular (a `load-entities` batch is atomic).

### Open questions for maintainers

1. **Codec: EDN-lines (as implemented) vs. keep CBOR and patch its float encoder.**
   EDN-lines makes type-exactness/determinism/closed-reader safety clean and drops
   the `clj-cbor` write dependency; CBOR is more compact at scale. Implemented
   behind a codec seam so CBOR can remain a variant. **This is the main thing to
   settle.**
2. Chunked-directory format acceptable, or single-file only?
3. Right home — `datahike.migrate`, or under the Wanderung umbrella for 1.0?
4. Target branch — `main` vs `development`.

---
*Data protection note:* a `:history? true` export resurrects retracted data
(retraction is not erasure in an immutable db). Use `:history? false` for data
leaving a trust boundary; the manifest records which kind it is.


alekcz:
Closing to validate locally first — will reopen once I've tested it end-to-end. The branch (feat/robust-export-import) stays up.

Issue #552, opened 2022-07-14
[Bug]:  "0.5.1506"  cannot import dump from "0.4.1480"


awb99:
### What version of Datahike are you using?

 "0.5.1506" 

### What version of Java are you using?

java 17.0.3

### What operating system are you using?

guix

### What database EDN configuration are you using?

{:store {:backend :file
:path "data/datahike-db"}
:keep-history? false}

### Describe the bug

Trying to import a db-dump made with  "0.4.1480" gets the following error:
The only difference in my project is the change of the datahiek version number.

```
Preparing import of data/datahike-dump/eavt-dump in batches of 1000
Execution error (ExceptionInfo) at clj-cbor.error/codec-exception! (error.clj:8).
Input data ended while parsing a CBOR value.


{:clojure.main/message
 "Execution error (ExceptionInfo) at clj-cbor.error/codec-exception! (error.clj:8).\nInput data ended while parsing a CBOR value.\n",
 :clojure.main/triage
 {:clojure.error/class clojure.lang.ExceptionInfo,
  :clojure.error/line 8,
  :clojure.error/cause "Input data ended while parsing a CBOR value.",
  :clojure.error/symbol clj-cbor.error/codec-exception!,
  :clojure.error/source "error.clj",
  :clojure.error/phase :execution},
 :clojure.main/trace
 {:via
  [{:type clojure.lang.ExceptionInfo,
    :message "Input data ended while parsing a CBOR value.",
    :data {:header 91, :cbor/error :clj-cbor.codec/end-of-input},
    :at
    [clj_cbor.error$codec_exception_BANG_
     invokeStatic
     "error.clj"
     8]}],
  :trace
  [[clj_cbor.error$codec_exception_BANG_ invokeStatic "error.clj" 8]
   [clj_cbor.error$codec_exception_BANG_ invoke "error.clj" 5]
   [clj_cbor.core$try_read_value invokeStatic "core.clj" 193]
   [clj_cbor.core$try_read_value invoke "core.clj" 185]
   [clj_cbor.core$decode invokeStatic "core.clj" 219]
   [clj_cbor.core$decode invoke "core.clj" 199]
   [clj_cbor.core$decode_seq$read_data_BANG___46451
    invoke
    "core.clj"
    236]
   [clojure.core$repeatedly$fn__6479 invoke "core.clj" 5138]
   [clojure.lang.LazySeq sval "LazySeq.java" 42]
[...]

Issue #262, opened 2020-12-04
Export needs sorting schema first to be importable


markusalbertgraf:
Hi,
when schema is added later in the life of a db, exported data can not be imported again without sorting schema to the top of the file first.
There is a script at [https://markusgraf.net/2020-12-03-Datahike-export-schema-sort.html](https://markusgraf.net/2020-12-03-Datahike-export-schema-sort.html) that does this as a short term solution along with some thoughts on what I think is going on.

markusalbertgraf:
I think Datomic solves this with partitions.
Partitions are:
- :db.part/db
- :db.part/tx
- :db.part/user
1) db/id are Longs
2) schema entries start at db/id 0
3) tx entries starting at db/id hex 0c 00 00 00 03 f0
4) data entries start at db/id hex 10 00 00 00 03 eb

In a new database there have already been a few transactions at creation time to preload schema.
Hex 3e8 equals decimal 1000.  So there seems to be an offset where tx and data counters start to reserve space for preloading and "hardcoding" of Schema.

mainej:
I'll second this. After exporting my db, I can't import it. The sequence of steps to reproduce the issue:

1. Transact schema for an attribute
2. Transact an entity with that attribute
3. Transact 1000 more datoms
4. Transact schema for a second attribute
5. Transact a value for the second attribute onto the entity created in *step 2*
6. Export
7. Import

This fails because the export happens in :eavt order. The entity has a lower :db/id than the attribute, so the data for the entity, including the values for both the first and second attribute, is exported first. Subsequently, the schema definition for the second attribute is exported. Then the import (which transacts in batches of 1000) fails because the attribute is used before it is defined.

mainej:
FWIW, here's some import code which seems to work with the format created by `datahike.migrate/export-db`, though it is slow:

```clojure
(let [conn (make-conn)]
  (->> (io/reader "/tmp/eavt-dump")
       line-seq
       (map read-string)
       ;; The existence of the extra Datoms to hold :db/txInstant throws of the
       ;; ordering of the entity creation, and causes errors
       (remove (comp #{:db/txInstant} :a))
       (sort-by :tx)
       (partition-by :tx)
       ;; Re-create database, one transaction at a time. This is slow, but
       ;; safe, because attributes are created before they are used, and
       ;; entities are created before they are referred to.
       (map #(d/transact conn %1))
       doall))
```

PR #922, opened 2026-07-30
feat(migrate): robust, type-exact, verifiable, streaming export/import (rebased from @alekcz)

whilo:
Rebases [**alekcz/datahike#5**](https://github.com/alekcz/datahike/pull/5) onto current `main` so it can be reviewed and CI-checked here. **All implementation and design work is by [@alekcz](https://github.com/alekcz) (Alex Oloo)** — his two commits are preserved with original authorship; anything we add lands as separate commits on top with `Co-authored-by` trailers.

It came out of a real ~98M-datom production migration that the current exporter could not survive.

## Why rebase rather than review in place

His PR was opened against his fork's `main` at `e4e26c68` (#905), which is 8 commits behind. That made the diff show `query.cljc` +268/−83, `lower.cljke` +150/−1, `execute.cljc`, `relation.cljc` and four query test files — **all of which is already-merged upstream work** (#908–#917), carried along as artifact. Reviewing it linearly would have wasted real effort, and merging as-is risked reverting recent query-engine fixes.

Against current `main` his actual change is **12 files, +2880/−360, zero query-engine files**.

## Verified on this branch, against current main

- his migrate suite: **20 tests, 78 assertions, 0 failures**
- full `clj-pss`: **862 tests, 7207 assertions, 0 failures**
- `bb format` clean

## What it does

Reworks `datahike.migrate` from the flat CBOR snapshot — marked *"temporary solution, pending developments in Wanderung"* in the source — into a full-history, type-exact, verifiable, bounded-memory dump & restore targeting local disk, any konserve store (S3 / S3-compatible / JDBC), or a diskless container.

| issue | fix |
|---|---|
| #633 | `:db.type/double` no longer round-trips as `Float`; values encode by runtime class, every builtin type incl. `float-array`/`double-array`/tuple |
| #377 | full history — asserts + retracts + tx entities; `history`/`as-of` match at every `t` |
| #262 | schema replays before the data that uses it |
| #508 / #531 | attribute-refs dumps translate refs to system entities by ident, never re-inserted |
| #552 | dumps carry a format version; old CBOR dumps still import via the legacy path |

Bounded memory at any scale (external merge sort out, streaming tx-aligned import in — validated 1.2 GB store → 285 MB dump → re-imported and verified under a **144 MB heap**), tiered verification, `estimate-import-memory`, and a `:sort? false` zero-scratch mode.

## Read `doc/import-export-design.md` §2 first

Its §2 grounds a prior spec against live source and **invalidates several load-bearing parts of it**. Two are worth calling out because they are the kind of thing a fresh implementation would get wrong:

- **The sort key must not put retractions first.** The direct transactor is order-sensitive and does not reorder within a tx: a card-one *add* upserts (`transaction.cljc:524`), and a *retract* matches an exact `[e a v]` and is a **no-op if that value was already replaced** (`:426`, `upsert?` at `:1511`). Retract-before-add — the intuitive choice — therefore silently drops a same-tx replacement's retraction.
- **Import is not resumable and cannot be**: the `:migration` id-map is not durable (`writing.cljc:56-58,227-281`), so it lives only on the in-memory db value. Honest recovery is recreate-and-restart, and `import-db` refuses a non-empty target. *Export* is resumable, because chunks are content-addressed.

It also correctly drops two things that prior spec demanded — a `read-string` RCE fix (the importer uses `clj-cbor`; no live vuln) and a core change for forward refs (`transact-entities-directly` already allocates eids at `transaction.cljc:1499-1500`).

## Known gaps, to be addressed as separate commits here

1. **The dump is not self-contained for `store-ref`.** `store-ref` is covered as a *value type* (rides as `#uuid`) but the referent **bytes are not carried**, so restoring a database that uses blobs yields datoms pointing at objects absent from the target store. `gc/reachable-store-refs` is the enumeration hook, and `gc/record-store-refs` already solved the identical ordering constraint for the konserve-sync walker (blobs must arrive *before* the reference that names them — the same discipline as manifest-last).
2. **Provenance + capabilities.** Carry `datahike.tools/meta-data` (the map `connector/version-check` already enforces for stores) **plus** a `:requires` capability set. A bare version is too blunt for a dump: a v3 dump using no v3-only features is readable by v2, and refusing it would work against datahike's backwards-compat stance. Capabilities make **downgrade** well-defined — fail precisely, never silently drop.
3. **Blob verification tier.** Blobs should verify at the same tier as datom chunks, so a dump cannot verify OK with a short `store-refs/`.
4. **Codec** — his one deliberately open question. Leaning canonical CBOR for the datom stream (RFC 8949, readers in every language — the scenario a dump exists for is the one where datahike is unavailable), EDN manifest, raw bytes for blobs.

## Where this sits in the migration story

Four axes, and this is the one that was missing:

| tool | axis |
|---|---|
| **konserve-sync** | physical, online, incremental replication (format-coupled) |
| **pg-datahike** | relational interchange, both directions incl. `pg_dump` / `COPY` |
| **wanderung** | Datomic (datom-shaped, lossy) — incomplete; harvest its adapter knowledge |
| **this** | datahike's own format: faithful, offline, self-contained, version-declaring |

None substitutes for another for structural reasons: konserve-sync ships the index format so it cannot cross an incompatible version; a logical dump re-derives ids so it cannot give cheap incremental replication.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -88,6 +88,9 @@
                                org.replikativ/kabel          {:mvn/version "0.3.100"}
                                is.simm/distributed-scope     {:mvn/version "0.1.6"}
                                org.replikativ/konserve-sync  {:mvn/version "0.1.35"}
+
+                               ;; S3-compatible store integration test (migrate-s3-test, needs docker; runs Garage)
+                               org.replikativ/konserve-s3    {:mvn/version "0.1.35"}
                                http-kit/http-kit             {:mvn/version "2.8.1"}}}
 
            :datomic {:extra-deps {com.datomic/datomic-free {:mvn/version "0.9.5703"}}}
--- a/dev/migrate_scale.clj
+++ b/dev/migrate_scale.clj
@@ -1,0 +1,81 @@
+(ns migrate-scale
+  "Manual scale harness for datahike.migrate: build a file-backed database larger
+   than the JVM heap, then export and re-import it under that same capped heap to
+   demonstrate that export/import memory is bounded by the buffers, not the db size.
+
+   Run with a small heap so the target is reachable quickly, e.g.:
+
+     clojure -J-Xmx140m -M:dev -e \"(load-file \\\"dev/migrate_scale.clj\\\") (migrate-scale/run)\"
+
+   `run` grows the db until the (estimated) EDN dump is at least `:heap-multiple`x
+   the max heap (default 2x), exports (chunked), and imports into a fresh store —
+   printing sizes and timings. NOT a unit test; it writes GBs to /tmp."
+  (:require [datahike.api :as d]
+            [datahike.migrate :as m]
+            [clojure.java.io :as io]))
+
+(defn dir-size ^long [path]
+  (->> (file-seq (io/file path)) (filter #(.isFile ^java.io.File %))
+       (map #(.length ^java.io.File %)) (reduce + 0)))
+
+(defn- mb [b] (quot (long b) 1048576))
+
+(defn run
+  "Options:
+     :heap-multiple  2      dump must be >= this many x the max heap
+     :value-chars    1000   size of the padding string per entity (bigger = fewer
+                            datoms to reach the target = faster build)
+     :store-cache    32     :store-cache-size (small: the node cache is count-bounded)
+     :sort-buffer    5000   export run size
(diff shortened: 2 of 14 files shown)


whilo:
This was implemented with #930 . Thanks @alekcz !

PR #933, opened 2026-08-02
feat(writer): :sync-commit? false — acknowledge after in-memory apply (opt-in)

whilo:
Tranche 3 of the pg-datahike campaign landing plan. Opt-in writer mode: with `:sync-commit? false`, transact acks after the in-memory apply; the store commit proceeds asynchronously in batches. Default (`true`) behavior unchanged — the flag simply isn't read.

Motivation: PostgreSQL-style OLTP workloads measure ack latency; on a memory store the durable-commit wait dominated small transactions. With this mode (plus the rest of the campaign) pgbench tpcb on the pgwire surrogate reaches parity with disk-PostgreSQL at 4 clients and exceeds it at 8.

Includes the commit-chain fix found in an adversarial review of the branch: the commit loop now threads the just-committed id as the next commit's parent instead of re-reading the connection (which, in async mode, holds a processed-but-uncommitted db whose meta can carry a stale id — mis-threading the chain on durable stores).

Draft because two ack-contract gaps are documented but not yet tested: the acked report lacks the final :db/commitId, and an async commit failure closes the writer after callers were acked. Tests for both (and a batching test) to follow before undraft. writer-error + api test namespaces green; full suites green within the campaign stack.


whilo:
Withdrawing after a controlled A/B on the final campaign stack: with commit-wait-time already 0, the writer's pipelining swallows the memory-store commit cost — sync mode reproduces the same pgbench results (tpcb c1 95 / c4 303 / c8 361 vs async 105-133 / 277-374 / 479, within run variance). The feature only reorders ack vs an in-RAM step on :memory stores, and on durable backends (where the latency win would be real) it trades away durability-on-ack — that deserves a deliberate design discussion (PG synchronous_commit=off analogue), not a side flag. The genuine writer opportunity is group-commit batching (+50-80% in isolated experiments, ~35k tps direct-PSS ceiling), which will come as its own design-first proposal.

PR #942, opened 2026-08-03
Re-transacting an entity spec no longer rejects a valid transaction

whilo:
**Stacked on #940** — review that first; this PR's diff against it is one function and two test files.

Applying your schema on every startup is an ordinary, deliberately idempotent pattern. It works twice and then fails forever:

```
transact #1  ->  :db.entity/attrs [:name]
transact #2  ->  :db.entity/attrs [:name :name]   (committed, no error)
transact #3  ->  THROWS :transact/schema, :invalid-updates
                 #:db.entity{:attrs [[:name :name] [:name]]}
```

## Mechanism

`di/-insert` is idempotent on `[e a v]` — asserting a datom the tree already holds returns the receiver untouched. But `with-datom`'s assert branch ran its whole accumulator chain regardless, so a redundant assertion moved every **derived** quantity while the index stood still.

`update-schema` `conj`s into the many-valued schema attributes (`:db.entity/attrs`, `:db.entity/preds`, `:db.attr/preds`), so each redundant assertion appends another copy. The schema-update check then compares what you are declaring against the shape the previous run wrote, and refuses.

## Severity

The mangled value is **persisted**, so the failure survives a reconnect — run 4 in a fresh connection is rejected too.

Nothing else is damaged. The indexes are correct, queries are unaffected, and the mangled spec still enforces, since `[:name :name]` demands exactly what `[:name]` does:

```
does the spec still ENFORCE? entity missing :name => rejected
does a VALID entity still pass?                   => accepted
```

But re-declaring the shape it now has is accepted and **doubles it again** — `[:name :name]` becomes `[:name :name :name :name]` — so there is no fixpoint to converge on. Recovery meant retracting the schema datom rather than re-declaring it.

## Why card-many only

Card-one goes through `with-datom-upsert` (`transact-add`'s `upsert?` is `(not (multival? db a))`), retraction is guarded by its own `if-some`, and purge by `current?`. The damage needs an attribute that is card-many **and** a schema attribute — which is what an entity spec is. Ordinary card-many user data hit only `:hash` and a phantom `:tx-data` entry, which is why this survived since the 2019 assert branch.

## The fix

Read the index count around the insert and gate only the non-idempotent work. `:aevt`/`:avet` stay unconditional: they are idempotent no-ops here, and running them lets an index that has skewed from `:eavt` self-heal. The count is read **before** the insert because the transaction loop runs over a transient index, where `-insert` mutates in place and both reads would see the same object.

Cost is two `-count` reads — O(1) on persistent-sorted-set — against a ~10 µs/datom pipeline. An A/B over six alternating rounds put the difference inside the harness's own ±3.5% noise floor.

Also closes the third `:hash` drift left open by #940.

## Not addressed here

`:tx-data` still reports a datom that changed nothing. DataScript searches first and routes redundant assertions to a key it drops before returning the report, so callers never see them. Changing that is public-behaviour surgery and wants its own decision.

## Verification

New `redundant-assertion-test`; with `db-hash-test`, 33 tests / 189 assertions, verified non-vacuous — **27 failures and 3 errors without the fix**, the errors being the schema rejection itself. Full suite **2668 tests / 33550 assertions / 0 failures**.

### DIFF
--- a/src/datahike/db/transaction.cljc
+++ b/src/datahike/db/transaction.cljc
@@ -412,16 +412,39 @@
       (log/raise "Attribute " a-ident " is :db.secondary/only but no secondary index covers it — its value would be lost"
                  {:error :transact/secondary-only-uncovered :attribute a-ident :datom datom}))
     (if (datom-added datom)
-      (cond-> db
-        true (update-in [:eavt] #(di/-insert % prim :eavt op-count))
-        true (update-in [:aevt] #(di/-insert % prim :aevt op-count))
-        indexing? (update-in [:avet] #(di/-insert % prim :avet op-count))
-        has-secondary? (update-secondary-indices a-ident datom true)
-        true (advance-max-eid (.-e datom))
-        true (update :hash + (hash prim))
-        schema? (-> (update-schema datom)
-                    update-rschema)
-        true (update :op-count inc))
+      ;; `-insert` is idempotent on [e a v] — asserting a datom the tree already
+      ;; holds returns the receiver untouched. Everything ACCUMULATING alongside
+      ;; it was not, so a redundant assertion moved the derived state while the
+      ;; index stood still. Worst of these is `update-schema`, which `conj`s into
+      ;; the many-valued schema attributes: re-transacting an identical entity
+      ;; spec grew `:db.entity/attrs` from [:name] to [:name :name], and the
+      ;; THIRD such transaction was then refused against the shape the second one
+      ;; wrote — an idempotent schema install failing permanently on boot 3.
+      ;;
+      ;; The count is read BEFORE the insert, not from `(:eavt db)` afterwards:
+      ;; the transaction loop runs over a TRANSIENT index, where `-insert`
+      ;; mutates in place and both reads would see the same object.
+      (let [eavt (:eavt db)
+            n-before (long (di/-count eavt))
+            eavt' (di/-insert eavt prim :eavt op-count)
+            inserted? (> (long (di/-count eavt')) n-before)]
+        (cond-> db
+          true (assoc :eavt eavt')
+          ;; `:aevt`/`:avet` stay UNCONDITIONAL. They are idempotent no-ops in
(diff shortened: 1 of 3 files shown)


whilo:
Superseded by #946 (`08c2ad81`), which fixes this and carries the tests. Verified on current main — five idempotent installs of the same entity spec:

```
transact #1 -> :ok   :db.entity/attrs [:name]
transact #2 -> :ok   :db.entity/attrs [:name]
transact #3 -> :ok   :db.entity/attrs [:name]
transact #4 -> :ok   :db.entity/attrs [:name]
transact #5 -> :ok   :db.entity/attrs [:name]
spec still enforces? yes — an entity missing :name is rejected
```

The analysis in this PR was right and carried over intact — `update-schema` `conj`ing into the many-valued schema attributes, the persisted mangling, no fixpoint to converge on, and why it needs an attribute that is card-many *and* a schema attribute. `redundant_assertion_test/an-idempotent-schema-install-stays-idempotent` is that scenario.

**What changed is the mechanism**, and it is worth recording why.

This PR gated on the index count around the insert. That is unsafe on a durable store — `PersistentSortedSet.count()` recomputes the count by walking the tree when it is not materialised:

```java
public int count() {
  if (_count < 0) _count = root().count(_storage);
  return _count;
}
```

`root().count(_storage)` takes the storage, so on a lazily-loaded index this faults nodes in — per asserted datom. Fine in the in-memory tests this PR was measured against, potentially very expensive on a file or S3 store.

#946 instead probes with `di/-slice` for the exact `[e a v]`, decided in `transact-add` before any index is touched. That also fixes two things this PR left alone:

- cardinality-one re-assertions re-dated the fact to the new transaction, leaving the current index and history disagreeing about when it was asserted;
- the `:tx-data` half — split out at the time as #944, and closed as fixed.

Scope also widened: the supersession retraction on a cardinality-one overwrite is now reported, since datahike already writes it into history and was not reporting it.

Thanks — the diagnosis here is what made the eventual fix tractable.

PR #957, opened 2026-08-16
Warm an index breadth-first, so a cold reader stops paying one round trip per node (EXPERIMENTAL)

whilo:
A cold reader's wall time over object storage is `misses × RTT` with **nothing overlapping**. Measured at +20 ms injected latency: the marginal cost of one more node was **24.96 ms** against a ~25 ms round trip. One node, one round trip, zero overlap — a scan asks for a node, blocks on the GET, and only *then* learns the next address.

It does not have to be that way, and the fix needs no prediction: a `Branch` holds `addresses()` — **every** child address — the moment it is materialized. So a whole level's addresses are known one level in advance. This walks breadth-first and fetches each level concurrently.

```clojure
(d/warm-index! db :eavt {:depth :interior :budget 2000})
(d/warm-datoms! db :avet [:issue/state] {})   ; warm exactly what this scan will read
(d/warm-db!    db {:depth :with-leaves})      ; all indices, one shared budget
```

## Why breadth-first, specifically

An iterator's walk is depth-first: each fetch depends on the previous one, which *is* the serialization being fixed. Breadth-first inverts it — a level's addresses are all known simultaneously, so the level goes out in one concurrent wave. Two further properties depend on it: `:interior` ("stop before the leaves") is only expressible level-by-level, and a partial budget is spent on the **shallowest, most-shared** nodes first rather than exhausting down one branch.

The two are complementary: the warm populates the cache breadth-first, then the ordinary in-order iterator runs against a warm cache and never blocks.

## Design properties

- **Two bounds.** `:depth` bounds shape, `:budget` bounds cost, whichever binds first wins. `:interior` expands while `level >= 2`, stopping exactly at the leaf boundary — exact, not heuristic, because leaves are level 0.
- **No modes.** `:with-leaves` with a budget above the tree simply runs out of frontier. "Preload everything" is not a separate code path — which is what makes a partial budget buy a *proportional* partial saving rather than falling off a cliff.
- **Round-robin across indices**, so `:eavt` cannot consume the whole budget before `:avet` sees any.
- **`warm-datoms!`/`warm-seek!` build bounds with `db.utils/components->pattern`** — the same call `db/contextual-datoms` makes. This matters: the pattern builder permutes components per index (`:avet` reads `[a v e tx]` → `datom(v,a,e,tx)`) and resolves idents, so hand-built bounds silently warm the *wrong* subtree.
- **Budget is clamped to `:store-cache-size`**, which is entry-counted — warming past it fetches nodes only to evict them.

## Validation

Unit tests: 16 tests / 53 assertions, green under both `*default-index*` bindings. **Full JVM suite 3163 tests, 0 failures.** ClojureScript suite 257 tests / 1553 assertions, 0 failures. Format clean; schemas verified under `malli.instrument/instrument!`.

Also validated **by consumption** — driven from a separate downstream project (a db-per-tenant SaaS template) against MinIO at +20 ms, i.e. as an external user rather than through datahike's own tests:

| tenant | depth | budget | warm GETs | query GETs |
|---|---|---|--:|--:|
| 400 issues | `:with-leaves` | 2000 | 37 | **0** |
| 400 issues | `:with-leaves` | 8 | 8 | 16 (naive: 20) |
| 400 issues | `:interior` | 2000 | 0 | 20 |
| 100 issues | `:with-leaves` | 2000 | 9 | **0** |

A full budget takes the query to **zero** object-store reads; a partial budget gives a partial saving strictly between naive and fully-warmed; `:interior` correctly fetches nothing on a height-1 tree, where the interior *is* the root that `:fuse-index-roots?` already delivers. Those are the properties — the exact counts vary with tenant shape and should be read as illustration.

Run against the same seeded tenant in separate JVMs, a prototype implementation and this API produce byte-identical `:fetched`, `:by-level`, `:by-index`, `:budget-left` and `:budget-exhausted?`.

## Intended home is persistent-sorted-set

The walk needs nothing datahike-specific — it is root → `addresses()` → children over `IStorage`. **Landing here first to get it measured and to unblock the object-store cold-read case**; extraction to PSS is planned, after which datahike keeps only the adapter (`components->pattern` bounds, the cache clamp, multi-index round-robin, the API surface), and other PSS consumers get the same fix for free.

Reviewers are therefore *not* being asked to bless this as a permanent home.

## The ClojureScript arm is a TODO, deliberately

`-warm!` returns `:unsupported :cljs` rather than throwing (a throw from an optional prefetch would turn a missing optimisation into a failed read) and rather than a silent empty report. Three `TODO(cljs)` markers name what is missing.

Its one real blocker — `child-bounds` needs PSS's private `binary-search-l` — **disappears in PSS**, so writing it here would mean writing it twice, in the harder place. The walk, budget, round-robin, depth policy and report are all shared `.cljc`, so it is a fill-in rather than a rewrite.

## Review questions

1. **`:siblings` on `-warm!`.** `-warm!` dispatches on one index but `warm-db!` needs one budget interleaved across several; slicing per-index would lose "spent exactly", which the no-cliff property rests on. So `-warm!` takes sibling indices, and the dispatch argument is also the first walked tree. This is the one awkward seam — and it gets simpler on both sides once the walk moves to PSS.
2. **`:supports-remote? false` is invisible at the call site.** A consumer on a `:datahike-server` writer gets silence; *whose* cache is warmed is a docstring-only contract today. A runtime warning may be worth adding.
3. `:sync?` is documented on `warm-index!` only, though all four entry points accept it.

## Not included

`bb codegen-clj-kondo` and `bb codegen-python` want to rewrite files with a large diff of **pre-existing** staleness (`valid-*`, `merge-db!`, `rseek-datoms`, `explain`, `branch!` are already missing from the kondo config). Reverted to keep this diff focused; worth a separate housekeeping pass.


whilo:
Trimmed the public surface from four functions to three (a36d558d), for the reason that this is experimental: adding an entry point later is cheap, removing one is breaking, so four was more than the feature needs to justify.

**`warm-seek!` is gone**, folded into `warm-datoms!` as `:unbounded?`. The two were the same call differing only in the upper bound, and the docstring now states the mapping in the form someone would search for:

```clojure
(warm-datoms! db idx cs)                    ;; <-> (d/datoms db idx & cs)
(warm-datoms! db idx cs {:unbounded? true}) ;; <-> (d/seek-datoms db idx & cs)
```

**`warm-index!` loses `:from`/`:to`** and becomes "warm this whole index, bounded by `:depth`/`:budget`". Range scoping now has exactly one entry point and it is the safe one. Hand-built datom bounds warm the *wrong* subtree silently: `components->pattern` permutes components per index (`:avet` reads `[a v e tx]` and produces `datom(v, a, e, tx)`) and resolves idents, so a bound built by hand in the wrong order is a valid datom naming a different part of the tree — no error, no wrong answer, just a warm that misses and a query that quietly pays full price. Nothing short of a coverage check notices. Shipping the trap beside the fix is worse than shipping only the fix, so the options are *dropped* (one chokepoint in `base-opts`) rather than honoured: a caller who passes them gets the whole index, never a silently wrong sliver.

**`warm-db!` is unchanged** — the common case, the round-robin budget sharing, and its `:indices` already covers "just these one or two". `:by-index` and `:budget-clamped?` stay in the report; they are how a decaying warm becomes visible as a metric before it shows up in p99.

The walk in `datahike.index.persistent-set.warm` still supports key bounds — that is what `warm-datoms!` drives it with. What no longer exists is a public entry point that accepts them from a caller.

Test coverage is preserved rather than dropped:

- the `warm-seek!` cases moved onto the flag (readahead is budget-bounded; unbounded warms strictly more than the bounded range), plus a new zero-`:reads`-delta coverage assertion against `d/seek-datoms` that the old `warm-seek!` tests did not have;
- the `:from`/`:to` range tests moved onto `warm-datoms!` — `[:item/id]` vs `[:item/id 300]` on `:avet` for wide-vs-narrow, and a fully specified `[e a v tx]` prefix for the point range (all four components make `components->pattern` fill nothing, so the two bounds are the same datom);
- two new tests pin the trim itself: `warm-index!` with stray `:from`/`:to` fetches exactly as much as without them, and `datahike.warm` publishes exactly three `warm-*` vars.

Also removed from the pod, CLI, native and Python exclusion lists and the `js-skip-list`. `datahike.api.types/SWarmIndex` is shared by the three that remain, so it stays as is — there was no `warm-seek!`-specific schema to delete.

Verification: full JVM suite 3175 tests / 24 342 + 11 410 + 102 assertions, 0 failures (3163 before; the delta is the new warm tests). `bb node-cljs-test` 257 tests, 0 failures. `shadow-cljs compile node-test` builds clean. `bb format` clean. `bb lint` adds no new class of error. Malli schemas still register and instrument — `malli.instrument/instrument!` wraps 51 vars including the three `warm-*`, and `datahike.api/warm-seek!` no longer resolves.

whilo:
Two follow-ups: the public names lose their bangs (53783804), and the tests stop filling `/tmp` (337a8902).

## `d/warm-index!` → `d/warm-index`, `warm-datoms!` → `warm-datoms`, `warm-db!` → `warm-db`

The fns in `datahike.warm` keep their bangs and become the `:impl` targets — the split `gc-storage` → `datahike.writer/gc-storage!` already uses.

The reason is sharper than "datahike's API has no bangs". In this specification a trailing `!` does not mean *side-effecting*, it marks the **async variant of a sibling**: `transact!`, `merge-db!`, `branch!` and `delete-branch!` all carry `:categories [… :async]`, while every synchronous op is unbanged however destructive — `gc-storage` deletes data. So `warm-index!` was claiming to be the async variant of a `warm-index` that does not exist.

That is not a style quibble, and the codegen had already noticed: `clj-name->java-method` maps a trailing `!` to an `Async` suffix, so the Java binding was generating `warmIndexAsync` for a call that is synchronous on the JVM. I had matched `branchAsync` and called it consistent; it was consistent with a name that means something else. The generator was reading the convention correctly — the name was wrong.

On top of that, warming changes no database state at all. The db value is immutable, results are identical warm or cold, and only latency and GET count move. What it mutates is a connection-scoped node cache, the same category as the read-through caching datahike already does invisibly.

**Confirmed Java names**, read out of `java/src-generated/DatahikeGenerated.java` (which `bb test` compiles, so this is checked rather than predicted):

```java
protected static final IFn warmDatomsFn = Clojure.var("datahike.api", "warm-datoms");
protected static final IFn warmDbFn     = Clojure.var("datahike.api", "warm-db");
protected static final IFn warmIndexFn  = Clojure.var("datahike.api", "warm-index");
```

`warmIndex` / `warmDatoms` / `warmDb`, no `Async`, no `warmSeek`.

One check worth naming, because it was the real risk in this rename rather than the public vars: **all four codegen exclusion maps now name ops that actually exist in the specification** — verified programmatically against `api-specification`, not by eye. A stale name in `pod`/`cli`/`native`/`python` excludes nothing and fails silently, so a half-done rename there would have quietly started generating FFI bindings for an operation that prefetches into a node cache the caller does not own.

## The tests were writing gigabytes to `/tmp`

Separate commit, and not a cosmetic one. `fresh-db` created a file-backed database per call and never deleted it — only the connection was released. This namespace alone had written **9.5 GB** into `/tmp` across two suite runs, which filled a 16 GB tmpfs and took the rest of the suite down with 56 errors and 2 failures, every one of them `No space left on device` in an unrelated namespace (scriptum, vectors, api-fs, store-connection). A failure mode that points everywhere except at its cause; I lost a full suite run to it before spotting that the leak was mine.

The size is intrinsic to the fixture, which is why the fix is cleanup rather than smaller data. These fixtures *must* be file-backed — the memory backend does not split the tree into per-node blobs, so a reconnected root reports zero child addresses and every assertion would pass vacuously — and they are deliberately 1200 items at branching factor 8 so the tree is deep enough for a multi-round walk. `fresh-db` now registers each config and an `:each` fixture deletes them after every test (`delete-database`, then the directory, which it does not necessarily remove), per-test so the peak is one test's databases rather than all of them.

## Verification (clean run, on the final tree)

- Full JVM suite: **2203 + 953 + 19 = 3175 tests, 35 860 assertions, 0 failures**, and `grep -c 'No space left'` = 0.
- Warm tests: 60 tests, 189 assertions, 0 failures, with **zero leftover `/tmp/warm-test-*` directories** and `/tmp` usage flat across the run.
- `bb node-cljs-test` 257 tests, 0 failures · `shadow-cljs compile node-test` clean · `bb format` clean · `bb lint` byte-identical to the previous commit.
- Malli: `instrument!` wraps 51 vars; `datahike.api` publics and spec ops are both exactly `warm-datoms`, `warm-db`, `warm-index`.

PR #977, opened 2026-08-22
Fence concurrent shared writers

whilo:
## Summary

- fence branch-head updates with Konserve revision tokens and conditional writes, retrying safe operations after head conflicts
- separate delivery from concurrency semantics: `:streaming?` continues to report whether a writer delivers completed writes into its connection, while `:writer-ownership` selects `:shared` or `:exclusive`
- default an omitted self-writer ownership to the safe `:shared` mode; `:exclusive` remains an explicit single-owner optimization, and existing databases can switch ownership without migration or metadata changes
- retain the experimental self-writer `:streaming?` config as a validated deprecated alias (`false` to `:shared`, `true` to `:exclusive`) without leaking self-only defaults into HTTP or Kabel writer configs
- conditionally claim database creation and fence branch creation, deletion, forced heads, and the shared branch registry so lifecycle operations cannot lose GC roots
- keep Konserve revision tokens as runtime-only state, outside stored database metadata and content-derived commit IDs
- expose and validate required fencing domains so multi-process deployments fail closed when the store cannot provide the requested guarantee
- preserve fatal writer failures for every caller and prevent retry queues from being stranded when a writer terminates
- skip empty immutable multi-assoc batches before writing the root/head
- use released org.replikativ/konserve 0.9.378 and refresh direct dependencies, including org.replikativ/kabel 0.3.108

## Validation

- persistent-set JVM profile: 1,023 tests / 12,922 assertions, all passing
- hitchhiker-tree JVM profile: 1,023 tests / 12,922 assertions, all passing
- ClojureScript Node suite: 270 tests / 1,748 assertions, all passing
- focused writer ownership/fencing suite: 105 tests / 411 assertions, all passing
- focused config, connection refresh, HTTP, and Kabel suites: 45 tests / 147 assertions, all passing
- formatting and diff checks pass
- lint task completes (repository-wide existing findings remain)
- antq reports no newer direct versions in deps.edn, bb.edn, or bb/deps.edn when unpublished local artifacts are ignored
- npm outdated reports no updates


whilo:
## Review round: fencing verified end-to-end, three findings fixed in 0a873a65

The core property holds: driven `dev/two_jvm_head_race.clj` across two real JVMs on a filestore — writer A parked in the window between its head read and head write, writer B committed, A's stale write was **rejected, retried against the moved head, and both transactions survived** (`{:names #{a b} :chain 3}`). The pre-PR failure mode was `b` silently lost. Connect-time `:require-fencing` checks are all correct, including on the cached-connection path.

Three findings, each reproduced before fixing and red-tested after:

**1. Multi-key stores fenced chained commits against a stale revision.** The multi-key path's fenced head write didn't pass `:with-revision? true` and never captured the revision it created, so the commit loop threaded the *pre*-commit stamp into the next chained group's fence. Measured from a **sole writer** on the memory backend: 24 manufactured head conflicts in 300 transactions with default retries (each a rejection + backoff + full re-apply), and with `:head-conflict-retries 0`, **295 of 300 transactions failed** — conflicts with no competing writer in existence. Affected every multi-key store (memory, jdbc, redis, dynamodb, indexeddb); the single-key path (s3, gcs, filestore) was always correct — so Lambda-on-S3 was fine, Lambda-on-DynamoDB was not.

**2. Legacy databases lost branch operations permanently.** An upgraded-in-place database has no `:revision` in its konserve metadata; reads hand back nil, and konserve refuses `:expected-revision nil` loudly by design. `commit!` handles this (first head write unconditional → key gains a revision → fenced thereafter — transact needs **zero** user action). But `update-branches!`/`assoc-current!` passed the nil through, and nothing else ever rewrites `:branches` — so `branch!`/`delete-branch!`/`force-branch!` failed forever, with no write that could heal it. Transacting first doesn't help; the regression test proves that too. Now both helpers fence when there's a token and write once unconditionally when there is none.

**3. `:require-fencing false` meant two things.** `check-fencing!` reads it by truthiness, `fencing-required?` read it with `some?` — so `false` (what nil becomes in JSON/env config) connected fine and then failed every commit. Both gates now read it as "not required". And `:datahike/fencing-unavailable` no longer kills the writer: retrying can't change a revisionless key, and a dead writer turns a migration condition into an outage. Verified end-to-end: repeated transacts keep answering the (now migration-explaining) error, one unfenced transact heals the head, and the demanding connection works from then on.

**Tests added**, both red without their fix:
- `a-sole-writer-never-manufactures-a-head-conflict` — 300 transacts, `:head-conflict-retries 0`, zero conflicts allowed. This is the assertion the suite was missing: retries self-heal manufactured conflicts into successes, so outcome-only checks stay green while every chained group burns a rejection.
- `a-legacy-database-heals-instead-of-refusing` — strips `:revision` from `:db` and `:branches`, then requires transact, `branch!`, and `delete-branch!` to all succeed.

Affected namespaces (writer-alternating, versioning-fencing, head-cache, config, background-gc, migrate-import-hostile): **79 tests, 294 assertions, 0 failures**. Two-JVM race re-run on the fixed branch: both writes survive.

**Migration story for users, verified:** transacting against an existing database just works — first commit is unfenced by design, everything after is fenced, no user action. Branch operations now heal the same way. The only deliberate refusal left is `:require-fencing` on a not-yet-healed head, and its error message now says exactly what to do.

PR #978, opened 2026-08-23
Extension-defined scalar value types

whilo:
## Summary

- add a conflict-checked, process-local registry for extension-defined scalar types
- dispatch schema validation and index comparison through exact dedicated wrapper types
- add versioned Fressian handlers for ordinary persistent-set stores
- fail reconnect clearly when the owning extension is absent, including store cleanup
- document equality, hashing, ordering, wire-version, and lifecycle contracts

This is an experimental foundation for consumers such as PostgreSQL JSONB. It deliberately does not yet claim attribute-ref databases, preconfigured/Boring stores, migration, Transit/remote CBOR, or custom values nested in typed tuples. Late registration and unsupported store capability guards remain follow-up work.

## Validation

- `clojure -M:format`
- `bb kaocha --focus datahike.test.value-type-test --focus datahike.test.schema-test` — 39 tests, 267 assertions, 0 failures across persistent-set, hitchhiker-tree, and specs
- `npx shadow-cljs compile node-test` — succeeds with the existing 24 warnings
- Node runner executes `datahike.test.value-type-test` first without failures before continuing into the broader suite
- `git diff --check`

The full Node suite is intentionally not treated as a gate for this PR; it has a pre-existing konserve Node filestore short-read failure in the background-GC area.


whilo:
Closing this experimental PR for now so pg-datahike can complete PostgreSQL semantics without committing Datahike to a public native-class/comparator extension contract. The branch and tests are intentionally preserved as design evidence. If we revisit extensible values, the likely direction is a Datahike-owned portable carrier whose persisted comparison key keeps databases traversable and exportable without plugin code, with serializer-specific support kept behind that logical representation.

PR #1025, opened 2026-08-28
deps: konserve 0.9.389 — close the target before the move, and the hardening behind it

whilo:
deps: konserve 0.9.389 — close the target before the move, and the hardening behind it

The second Windows fix. `create-database` failed on the first write:

    …\dh-test-store\03ac….ksv.new -> …\dh-test-store\03ac….ksv

`AccessDeniedException(source, target, null)` from `Files/move`: Windows
refuses to replace a file with an open handle, and konserve held the target
open — and locked — across the move. 0.9.389 releases both before the move.

The review of that change (six rounds, gpt-5.6-sol) found the release exposed
a shared staging path, and then kept going: per-writer staging names, `dissoc`
ordered by the sidecar fence, `k/revision` making the key fenceable, truncation
on shorter in-place writes, full writes with honest strides on JVM and Node,
Reader input encoded statefully. Final verdict: no remaining High or Medium.
konserve's suites: JVM 262 tests / 4302 assertions, Node 66 / 424, 0 failures.

Verified here: `bb format` and `bb reflection` clean; migrate suite result
posted on this PR. Windows CI is the real check.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -13,13 +13,17 @@
         ;; and hasch. Datahike compiles konserve into `dthk`, where a reflective
         ;; call resolves at run time and so works only if it was registered when
         ;; the image was built.
-        ;; 0.9.388: the file store no longer fails `delete-store` on Windows
-        ;; (konserve#186). It fsyncs the parent directory after removing an
-        ;; entry, Windows cannot open a directory as a channel, and the catch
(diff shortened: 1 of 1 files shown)


whilo:
Migrate suite against konserve 0.9.389 (`bb test migrate`, local): **277 tests, 1452 assertions, 0 failures.** Format and reflection gate clean.

PR #1026, opened 2026-08-28
chore: one convention for temporary directories, resolvable off POSIX

whilo:
chore: one convention for temporary directories, resolvable off POSIX

The test suite chose temporary store paths eleven different ways, and
seventy-eight of them were the literal `/tmp`. On Windows, Java resolves
`/tmp/x` to `\tmp\x` on the current drive, which does not exist — the same
shape as the native-image test failures fixed in #1014, applied across the
whole suite. Nothing here runs on Windows CI yet; this is what makes it
possible to.

## One convention

`datahike.migrate.fs/temp-dir!` already existed and was already right: it
resolves against `java.io.tmpdir` (`os.tmpdir()` on Node), builds the path
with the platform separator, and returns a unique directory. Every site now
goes through it — with one distinction the first pass got wrong and the second
commit corrects:

`temp-dir!` CREATES the directory, and konserve's `create-store` REFUSES a path
that already exists. That failed twenty-one tests, and hid in most of the
rest because they `delete-database` first, which removed the directory the
helper had just made. So a store gets `temp-store-path!` — a not-yet-existing
child of one temp root per process — and a dump or Lucene directory, which
must exist, keeps `temp-dir!`. The root-per-process is itself a fix from
review: one created parent per test was one leaked directory per test.

## Also

- Two `case (System/getProperty "os.name") "Windows 10" …` sniffs deleted:
  they matched only that literal, so `Windows Server 2022` fell through to
  `/tmp` anyway.
- Three duplicated `tmp-path` helpers collapsed into the shared one.
- `src-secondary/.../scriptum.clj` shipped `(str "/tmp/scriptum-" (random-uuid))`
  as the DEFAULT index path — library code, a user-facing Windows bug.
  Fixed, with its doc row.
- `bb test python` routed through `bash` like its siblings (a shebang, which
  Windows does not honour).
- `tools.reflection` derived namespace names with `/` and a `^root/` strip.
  On Windows `.getPath` yields backslashes, neither fires, and the gate
  reported "no reflective calls" having compiled nothing. Separator-agnostic
  now, an empty namespace list fails, and — from review — a namespace that
  fails to load fails the gate with the compiler's stderr instead of being
  silently skipped.
- A leaked agent-scratchpad path, `/tmp/claude-1000/…`, committed in #930 in
  two tests. Gone.

## Deliberately left alone, each annotated in place

- `migrate_test.clj`'s `/tmp/should-not-exist` and `migrate_init_build_test`'s
  `"/tmp"` — refused before the path is read.
- `config_test.cljc:27/33` — URI parsing, the output must spell the input.
- `backward_test.clj` — the WRITER runs the last release's frozen copy of the
  file, which spells `/tmp` literally; this side must match it. Moving it is a
  two-release step, and the comment says how.
- `depr_config_uri_test.clj` — the deprecated URI form does open a store at
  `/tmp`; making it portable means teaching `uri->config` Windows drive
  letters. Linux CI only. Its comment now says so, truthfully.

## Verified

External review (gpt-5.6-sol, medium): four findings, all addressed in the
third commit — the frozen-writer mismatch, the untruthful URI comment, the
gate swallowing load failures, the leaked parent directory. Cleared
explicitly: the temp-dir!/temp-store-path! split at every site, no test
needing a stable path broken, no assertion weakened, `.cljs` sites on Node
APIs, the scriptum default equivalent on Linux.

Rebased onto main after #1022/#1025/#974. `bb format` and `bb reflection`
clean on the rebased tree; `clj-pss` and `migrate` results are posted below.

### DIFF
--- a/bb/src/tools/reflection.clj
+++ b/bb/src/tools/reflection.clj
@@ -62,10 +62,19 @@
             (System/exit 1))
         form (str "(set! *warn-on-reflection* true)"
                   "(doseq [n '" (pr-str (vec (map symbol nses))) "]"
-                  "  (try (require n) (catch Throwable _ nil)))")
+                  "  (require n))")
         _ (println (format "Compiling %d namespaces with *warn-on-reflection*..."
                            (count nses)))
-        {:keys [err]} (p/sh "clojure" "-M:libdatahike" "-e" form)
+        {:keys [err exit]} (p/sh "clojure" "-M:libdatahike" "-e" form)
+        ;; A namespace that does not load is a hard failure, not a skip. The
+        ;; old form caught every Throwable, so a broken require was silently
+        ;; dropped and the gate reported "no reflective calls" over whatever
+        ;; happened to compile — the same silent pass the empty-list guard
+        ;; below exists to prevent, one level down.
+        _ (when-not (zero? exit)
+            (println "Compilation failed (exit" exit ") — the reflection gate did not run:")
+            (println err)
+            (System/exit 1))
         warnings (->> (str/split-lines (or err ""))
                       (filter #(str/starts-with? % "Reflection warning"))
                       (filter #(str/includes? % "datahike/"))
--- a/bb/src/tools/reflection.clj
+++ b/bb/src/tools/reflection.clj
@@ -16,6 +16,27 @@
 
 (def ^:private src-roots ["src" "libdatahike/src"])
 
+(defn- ns-name-of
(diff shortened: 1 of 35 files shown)


whilo:
On the rebased branch (after #1022, #974, #1025): `bb test clj-pss` **1091 tests, 13235 assertions, 0 failures**; `bb test migrate` **277 tests, 1452 assertions, 0 failures**. `bb format` and `bb reflection` clean.

PR #1029, opened 2026-08-28
Preserve scalar bind selectivity in query plans

whilo:
Separates physical scan work from logical output cardinality for local fused entity-group planning, and propagates scalar/tuple function-bind cardinality to direct consumers independent of clause order. This prevents SQL-style typed parameter binds from turning primary-key probes into full scans and lets selective unindexed filters drive their entity group. Scope is deliberately local: exported inter-group cardinality remains conservative. Validation: 255 tests, 24,333 assertions across planner, IR, engine parity, binding seam, query-shape corpus, and seeded differential suites. In the pg-datahike REPL at 100k rows, a prepared primary-key lookup fell from about 110 ms to 0.4 ms; a selective unindexed scan fell from about 59 ms to about 5 ms and scales linearly from 1k through 100k rows.

### DIFF
--- a/src/datahike/query/lower.cljc
+++ b/src/datahike/query/lower.cljc
@@ -905,13 +905,15 @@
      (plan/source-cards {:kind :pattern :classified (scan->classified node) :db db
                          :bound-var-cards bound-var-cards})
 
-     ;; Function binds whose var advertises :datahike/output-cardinality (e.g.
-     ;; a graph algorithm whose result size is data-dependent). Without this the
-     ;; bind is opaque and downstream patterns over its output var fall back to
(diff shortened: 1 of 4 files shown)


whilo:
Rebased over #1031/#1030 and fixed the three red test jobs. They all hit the same stale graph-planner assertion: algorithm output-cardinality metadata does not expand a scalar binding, but the scalar still emits one relation row per input row, so a downstream entity pattern is a one-value probe rather than an unbounded attribute scan. Local validation: full clj-pss (1,093 tests / 13,241 assertions), the exact graph-planner namespace under clj-pss, planner-disabled clj-pss, and clj-hht, plus query-planner/query-IR/query-shape corpus (64 tests / 225 assertions), all green.

PR #1057, opened 2026-09-03
Kabel listener in the server and a datahike/kabel npm entry

whilo:
First authenticated browser-to-server version over Kabel.

- Server: opt-in `:kabel` listener (`:host :port :peer-id :jwt :store`) serving `kabel.remote` and konserve-sync next to the HTTP routes; JWT validated by the server; remote calls and store subscriptions authorized by the server's `:authorize` policy, the same eacl relationships as the HTTP routes (dispatch = `:transact`, create/delete their own ops, subscription = `:read`; no policy = every authenticated caller, as over HTTP).
- npm: opt-in `datahike/kabel` entry: `createKabelPeer` (token as string or function, read at every connection), `connectKabelPeer`, `maintainKabelPeer` (reconnection with backoff and status events), `refreshKabelToken`, `invokeRemote`, `registerRemoteFn`, `stopKabelPeer`, with TypeScript types.
- The remote runtime is `kabel.remote` (kabel 0.3.129); distributed-scope leaves the dependencies and with it the ClojureScript compiler leaves the server JAR.

Landed along the way, in kabel: `kabel.remote` and reconnection/refresh (#28), quiet serve shutdown (#29), and the pub/sub frame-size bug that hung every Datahike handshake since kabel 0.3.113 (#30).

- [x] Authorization through the server's `:authorize` policy on both the remote-call and the sync-subscribe path, with tests (`datahike.test.http.kabel-test`).
- [x] Reconnection and re-authentication in the client.
- [x] distributed-scope dropped; the connector fails fast on a refused subscription.
- [ ] Measure the standalone JAR after the dependency change (expected well under the previous ~101 MB).
- [ ] Not in this PR, stated in the README: durable offline write queue, multi-tab coordination, re-subscribing store topics after a reconnect (the connector subscribes once per connect).

https://claude.ai/code/session_01J9RUNkHFidHP8EvQVCSqf3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ npm-package/*.d.ts
 npm-package/package.json
 npm-package/browser/
 npm-package/s3/
+npm-package/kabel/
 npm-package/LICENSE
 npm-package/karma.conf.js
 !npm-package/test.js
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,11 @@ When something is added, it's typically marked *Experimental*. When the API cont
   The remote-invocation runtime is `kabel.remote` (kabel 0.3.129);
   distributed-scope is no longer a dependency, which removes the ClojureScript
   compiler from the server JAR. The connector fails fast when a store
-  subscription is refused instead of waiting forever.
+  subscription is refused instead of waiting forever, and a write waits for
+  its own sync for at most two minutes. The listener reopens the databases an
+  earlier run created, so they survive a restart, and gates the string
+  spelling of its remote functions and the transaction-report topics as what
+  they are.
 
 - **The standalone server drains work on shutdown.** SIGTERM stops new accepts,
   waits up to 30 seconds for in-flight HTTP operations, then releases database
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ When something is added, it's typically marked *Experimental*. When the API cont
 
 ## 0.8
 
+- **Kabel listener and browser client.** The standalone server takes an
+  optional `:kabel` listener (JWT-authenticated WebSocket next to the HTTP
+  routes) and the npm package an opt-in `datahike/kabel` entry for a tiered
+  memory + IndexedDB replica with the `:kabel` writer. Remote calls and store
+  subscriptions are authorized by the server's `:authorize` policy, the same
+  eacl relationships as the HTTP routes. The client reads its token from a
+  function at every connection and refreshes it before expiry
+  (`refreshKabelToken`), reconnects with backoff (`maintainKabelPeer`), and
+  can call and serve remote functions (`invokeRemote`, `registerRemoteFn`).
+  The remote-invocation runtime is `kabel.remote` (kabel 0.3.129);
+  distributed-scope is no longer a dependency, which removes the ClojureScript
+  compiler from the server JAR. The connector fails fast when a store
+  subscription is refused instead of waiting forever.
+
 - **The standalone server drains work on shutdown.** SIGTERM stops new accepts,
   waits up to 30 seconds for in-flight HTTP operations, then releases database
   connections, permissions state, and metrics ownership exactly once. The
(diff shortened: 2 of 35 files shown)


whilo:
Standalone JAR measured on ea04084: 74.4 MB, down from about 101 MB. No `cljs/analyzer` or Closure compiler entries remain; the `cljs/` files left are core.async's own ClojureScript sources shipped in its jar.

whilo:
CI note: on ba21952 the kabel tier hung in `test-fork-branch-reflects-to-subscriber` (incremental block) until the 5-minute no-output limit, while the same path passed on the two commits before it and passes locally five times in a row, twice under CPU contention. The connect's handshake wait is now bounded (120 s, `:handshake-timeout-ms`), so a recurrence fails with the store and branch in the report instead of silently.

whilo:
Correctness and security review of the listener, handlers and writer (read-only agent, findings traced to code). Fixed in 4eecd943:

- **High, fixed:** databases created by an earlier run were not re-registered after a server restart, so dispatch and subscribe failed until the database was deleted. The listener now reopens every UUID-named directory below its store path with the stored configuration.
- **Medium, fixed:** `"datahike.kabel/dispatch"` spelled as a string resolved in kabel but bypassed the op mapping and was asked as a host `:invoke`; under a host policy granting `:invoke` to non-admins this let them run dispatch/create/delete on any store. Gate normalises to the symbol and refuses any other name under `datahike.kabel`. A store id that is not a UUID reaches no database.
- **Medium, fixed:** tx-report topic subscriptions were asked as a read of a database named by the topic keyword (admin-only under eacl); now asked as a read of the store behind it.
- **Medium, fixed:** the writer failed on `gc-storage!` and secondary-index results (no commit watermark); they are delivered directly. The wait for a write's own sync is bounded at two minutes.
- **Medium, fixed:** `start!`/`stop!` swallowed a failed peer start (port in use looked like a running listener).
- **Low, fixed:** client-supplied request ids must be UUIDs; debug printlns are log lines; the release helper's dead branch and non-Throwable error path.

Checked and fine: principal forgery (auth middleware stamps or strips before remote/pubsub; `serve` overwrites any `:kabel/principal` in the arg map), unauthenticated peers, JWT alg pinning and expiry, path traversal (konserve validates the UUID id before touching the filesystem), takeover of an existing store id, branch and op mapping parity with the HTTP gate.

Deferred, tracked for follow-up: `create-database` runs blocking `d/connect`/`release` inside go blocks with no cap on the store registry (admin-only under eacl today); delete races a transaction already inside the writer loop and does not tombstone subscribers; the store registry and remote registry are process-global, so `stop!` does not tear them down and two listeners in one JVM would share them; the head synced to clients carries the server's `:store :path`.

https://claude.ai/code/session_01J9RUNkHFidHP8EvQVCSqf3

whilo:
Second review (read-only agent over the whole branch), fixed in 8ea6246a:

- **Medium:** the writer discarded the shutdown error delivered on its wait channel and finalised a dispatch on a released connection; the error is delivered now.
- **Medium:** a connect that failed after opening the store leaked the subscription, the codec registration and the store (an IndexedDB handle in a browser); the failure path releases all three.
- **Medium:** the npm entry pushed nil results through `async/map`, which only works with assertions elided; nil now travels as a closed channel, yielded exceptions are freed through `<?`, and `refreshKabelToken` rejects instead of throwing synchronously when there is no live connection.
- **Low:** the create and delete handlers check the store id is a UUID themselves rather than relying on konserve; listener tests use free ports.
- **Docs:** grants shown with the shared token (the shown server config has no HTTP JWT validator), the policy example destructures `:fn-name`, stale "does not reconnect" sentences removed, a token source is a function returning a string or a promise.

Checked and found fine: middleware order and principal stamping, the gates' spelling and UUID handling, `:default` composition, the non-blocking server path, connector/writer bookkeeping, server lifecycle ordering, the JSON middleware change (no API route takes an object body; a map body threw before), npm exports and the TypeScript declaration test.

Still open, next in kabel: the permission decision itself (`eacl/can?`, a query on the system database) runs inside kabel's go blocks on the remote and subscribe gates. With a file-backed system database a cache miss is storage IO on a dispatch thread. The fix is a gate that may return a channel, so Datahike can decide on `async/thread`; that is a kabel change, then a datahike bump. Also still deferred from the first review: registry cap, delete ordering with a tombstone, per-peer registries.

https://claude.ai/code/session_01J9RUNkHFidHP8EvQVCSqf3

Issue #1024, opened 2026-08-28
[Bug]: Query planner: _ argument to a rule in an or → "Cannot resolve any more clauses"


mokshasoft:
### What version of Datahike are you using?

0.8.1831

### What version of Java are you using?

OpenJDK 25.0.2

### What operating system are you using?

MacOS

### What database EDN configuration are you using?

{:store {:backend :memory
           :id #uuid "00000000-0000-0000-0000-00000000ede4"}
   :schema-flexibility :write
   :keep-history true}

### Describe the bug

Invoking a Datalog rule with a blank `_` in one of its argument positions (inside an
`or`) crashes the compiled query planner: it compiles the `_` into an unresolvable
`(identity _)` clause instead of treating it as an anonymous variable, then throws
`clojure.lang.ExceptionInfo: "Cannot resolve any more clauses"`. The identical query
and database return the correct result on the legacy engine, and inlining the same
`or` without the rule also works under the planner — so the crash is specific to the
planner's handling of a rule call that has a `_` argument.

### What is the expected behaviour?

The planner should execute the query and return the same result as the legacy engine
(here, `#{}` — 0 rows). A `_` argument to a rule call should be treated as an
anonymous/unbound variable, not compiled into an unresolvable `(identity _)` clause.

### How can the behaviour be reproduced?

A self-contained, DB-free repro script (`repro_planner_rule_crash.clj`, ~1s) is here:
https://gitlab.com/arbetsformedlingen/taxonomy-dev/backend/jobtech-taxonomy-api/-/commit/bee48ebb55dc7e52c306b070a6ff911fa79b224e

```
clojure -Sdeps '{:deps {org.replikativ/datahike {:mvn/version "0.8.1831"}}}' -M repro_planner_rule_crash.clj on
```

`on` enables the planner (the crash); `off` (or `DATAHIKE_QUERY_PLANNER=false`) is the control and returns `#{}`. It throws `Cannot resolve any more clauses` with `ex-data {:error :query/where, :clauses [[(identity _) ?…]]}` at `datahike.query.execute/bind-by-fn-strict` (execute.cljc:2929).

PR #1065, opened 2026-09-04
Fix/cache db view hash

### LINKED ISSUE #1063: [Bug]: db-view-hash rescans every datom on each view hasheq/hashCode — O(n) per call, uncached

0.8.1831

OpenJDK 25

MacOS

{:store {:backend :memory :id #uuid "00000000-0000-0000-0000-0000000000fa"}
 :schema-flexibility :write
 :keep-history true}

### Describe the bug

The temporal/filtered DB views (`AsOfDB`/`HistoricalDB`/`SinceDB`/`FilteredDB`) hash
themselves via `db-view-hash`, which is
`(reduce #(+ %1 (hash %2)) 0 (dbi/datoms db :eavt []))` — a full walk over every
datom — and it's called directly from each view's `hasheq`/`hashCode` with **no
memoization**. A plain `DB` reads its precomputed `:hash` (O(1)); the views don't.

So **the same view instance re-scans all of its datoms on every hash**. Using a view
as a map/set key, or anything that hashes the same view value more than once, pays a
full datom scan each time — O(n) per call, growing with the view's datom count. A
batching/caching layer that keys on the db value (e.g. an `as-of` db used as a cache
key) hashes it repeatedly across a request, so a large history is rescanned many
times over — enough to stall a request on real data.

The content hashing itself is correct and intentional — it's what keeps a view and
an equal `DB` hash-consistent for `equiv-db` (#939). The issue is purely that it
isn't cached.

I will push a PR shortly.

Hashing a view instance should be O(1) after the first call (like a `DB`), one scan
to compute the sum, then a cached read.

Count datom scans (via `datahike.db.interface/datoms`, which `db-view-hash` calls)
while hashing **one** view instance five times — it should be 1, but it is 5:

```
clojure -Sdeps '{:deps {org.replikativ/datahike {:mvn/version "0.8.1831"}}}' -M -e '
(require (quote [datahike.api :as d]) (quote [datahike.db.interface :as dbi]))
(def cfg {:store {:backend :memory :id #uuid "00000000-0000-0000-0000-0000000000fa"}
          :schema-flexibility :write :keep-history true})
(d/delete-database cfg) (d/create-database cfg)
(def c (d/connect cfg))
(d/transact c [{:db/ident :k :db/valueType :db.type/string :db/unique :db.unique/identity :db/cardinality :db.cardinality/one}])
(d/transact c [{:k "a"} {:k "b"} {:k "c"}])
(def asof (d/as-of (d/db c) (:max-tx (d/db c))))
(def scans (atom 0))
(let [orig dbi/datoms]
  (with-redefs [dbi/datoms (fn [& a] (swap! scans inc) (apply orig a))]
    (dotimes [_ 5] (hash asof))))
(println "datom scans across 5 hashes of ONE as-of view:" @scans "(should be 1)")'
```
Output on `0.8.1831`:
```
datom scans across 5 hashes of ONE as-of view: 5 (should be 1)
```
Each hash re-scans; the count equals the number of hashes, and each scan is O(datoms)
— so the per-hash cost grows with the view's size. (A plain `DB` scans zero times
here — it reads its cached `:hash`.)

mokshasoft:
#### SUMMARY

Fixes #1063.

The DB views (`FilteredDB` / `HistoricalDB` / `AsOfDB` / `SinceDB`) hash themselves by
content via `db-view-hash` — a full O(n) datom sum — on **every** `hasheq`/`hashCode`
call, with no memoization. A plain `DB` reads its precomputed `:hash` (O(1)); the views
don't. So using a view as a map/set key re-scans all of its datoms on every lookup, and
a batching/caching layer that keys on an `as-of` db can rescan a large history many
times per request.

The content hashing itself is correct and intentional — #939 added it so a view and an
equal `DB` hash consistently for `equiv-db`. This PR only makes it cheap: each view
record gets a `hash-cache` field (a `volatile!`); `db-view-hash` computes the sum on the
first call and reads the cache afterwards. The first `hasheq`/`hashCode` scans; later
calls are O(1). The hash **value is unchanged**, so the `equiv-db` consistency invariant
is preserved.

Design decisions:
- **Per-instance, not a global cache** — no shared mutable state, no leak, GC-natural,
  and it mirrors how `DB` already carries `:hash`. Views are immutable once constructed
  (only ever `assoc`'d with non-content keys like `:system`/`:history?`), so the cached
  value never goes stale.
- The field is threaded through **all** view constructors, including the EDN readers, so
  nothing can build a view without a cache cell.
- `db-view-hash` takes the cache as a **parameter**, read by each caller from its own
  typed record via `(.-hash-cache db)`. This keeps the shared fn free of a polymorphic
  field access ClojureScript can't type-infer (`FilteredDB` blocks keyword/`valAt`
  lookup, so interop is the only cross-platform accessor).

##### Bugfix
- [x] Integration tests added  
- [x] Formatting checked  

#### ADDITIONAL INFORMATION

Adds three commits: a consistency test (`FilteredDB` mirrors a `DB` → equal + hash-equal;
two views over the same content hash-equal), a caching guard (deterministic scan count,
red on `main`), and the fix (green).

Validation:
- **clj** — `bb test` (`:persistent-set` + `:hitchhiker-tree` tiers, plus
  `migrate`/`norm`/`specs`; `migrate` exercises the readers this PR touches): 2826 tests /
  30119 assertions, no new failures. The only reds are pre-existing
  `datahike.test.http.server-test` flakiness (fail identically on `main`).
- **cljs** — `bb node-cljs-test`: 289 tests / 1823 assertions, 0 failures, 0 warnings.
- **format** — `bb format`: clean.
- **native-image / back-compat** — not exercised locally: both fail on `main` too in this
  environment for unrelated reasons (native-image trips a pre-existing
  `clojure.core.async.timers` "started Thread in the image heap" on GraalVM 24; back-compat
  fails building the cloned old version's Java codegen under JDK 26). Left to CI.

Before / after — the caching guard counts datom scans over five hashes of one view instance.

Without the fix (`db-hash-test`):

```
FAIL in (hashing-a-view-scans-its-datoms-once-then-serves-cached) (db_hash_test.cljc)
  HistoricalDB: expected one datom scan across five hashes, got 5
  AsOfDB:       expected one datom scan across five hashes, got 5
  SinceDB:      expected one datom scan across five hashes, got 5
  FilteredDB:   expected one datom scan across five hashes, got 5
Ran 7 tests containing 25 assertions.
4 failures, 0 errors.
```

With the fix:

```
Ran 7 tests containing 25 assertions.
0 failures, 0 errors.
```

A view instance now costs one datom scan no matter how many times it is hashed.

This PR, together with a yet unimplemented fix for #1024, closes the regressions we've found in our test suites.

### DIFF
--- a/src/datahike/api/impl.cljc
+++ b/src/datahike/api/impl.cljc
@@ -148,7 +148,7 @@
 
 (defn since [db time-point]
   (if (dbi/-temporal-index? db)
-    (SinceDB. db time-point)
+    (SinceDB. db time-point (volatile! nil))
     (log/raise "since is only allowed on temporal indexed databases." {:config (dbi/-config db)})))
 
 (defn as-of
@@ -177,10 +177,10 @@
   (if (dbi/-temporal-index? db)
     (if (int? time-point)
       (if (<= const/tx0 time-point)
-        (AsOfDB. db time-point)
+        (AsOfDB. db time-point (volatile! nil))
         (log/raise (str "Invalid transaction ID. Must be bigger than " const/tx0 ".")
                    {:time-point time-point}))
-      (AsOfDB. db time-point))
+      (AsOfDB. db time-point (volatile! nil)))
     (log/raise "as-of is only allowed on temporal indexed databases." {:config (dbi/-config db)})))
 
 (defn history [db]
@@ -191,7 +191,7 @@
     ;; legacy lookup path on cljs (the fused temporal path needs the origin's
     ;; :eavt to be a real PersistentSortedSet, which a nested wrapper isn't).
     (instance? HistoricalDB db) db
-    (dbi/-temporal-index? db) (HistoricalDB. db)
+    (dbi/-temporal-index? db) (HistoricalDB. db (volatile! nil))
     :else (log/raise "history is only allowed on temporal indexed databases." {:config (dbi/-config db)})))
 
 ;; `vt-meta-attrs`, `tx-covers-at?`, `find-eav-winner`, `mk-vt-pred`,
--- a/src/datahike/core.cljc
+++ b/src/datahike/core.cljc
@@ -120,8 +120,8 @@
      (let [^FilteredDB fdb db
            orig-pred (.-pred fdb)
            orig-db (.-unfiltered-db fdb)]
-       (FilteredDB. orig-db #(and (orig-pred %) (pred orig-db %))))
-     (FilteredDB. db #(pred db %)))
+       (FilteredDB. orig-db #(and (orig-pred %) (pred orig-db %)) (volatile! nil)))
+     (FilteredDB. db #(pred db %) (volatile! nil)))
    assoc :datahike/secondary-filter-provenance :opaque))
 
 ; Changing DB
--- a/src/datahike/db.cljc
+++ b/src/datahike/db.cljc
@@ -482,7 +482,7 @@
 
 ;; FilteredDB
 
-(defrecord-updatable FilteredDB [unfiltered-db pred]
+(defrecord-updatable FilteredDB [unfiltered-db pred hash-cache]
   #?@(:cljs
       [IEquiv (-equiv [db other] (equiv-db db other))
        ISeqable (-seq [db] (dbi/datoms db :eavt []))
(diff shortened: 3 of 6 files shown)


whilo (approved):
Thank you!

PR #1076, opened 2026-09-06
Test transaction failure propagation from directory-sync barriers

whilo:
## Summary
Add a JVM file-store regression test injecting AccessDeniedException and IOException at the Konserve directory-sync boundary after a successful seed transaction.

Verify that the barrier is reached, the transaction returns an error within a bounded time with the original storage cause, the connection does not publish the failed commit, and subsequent writes fail because the writer has stopped. Deliberately make no disk-rollback assertion: a failed barrier can follow a rename.

## Validation
- Against revised Konserve replikativ/konserve#190 (41c5a4b), using a local dependency override: 1 test, 10 assertions passed.
- Against the released dependencies declared by current Datahike main, together with writer-error-test: 4 tests, 23 assertions passed.
- Linux JVM tests only; no Windows or power-cut qualification.

No runtime or dependency changes. This covers Datahike error propagation, not whether a backend silently suppresses an error. Konserve #190 separately tests and fixes that boundary.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -27,7 +27,9 @@
         ;; 0.9.391 adds guarded, process-wide metrics sinks around API, blob I/O
         ;; and lock waits. With no sink installed its measurement sites take the
         ;; allocation-free fast path.
-        org.replikativ/konserve                     {:mvn/version "0.9.391"}
+        ;; 0.9.393 makes directory persistence fail closed and supplies the
+        ;; Windows native barrier (JDK 22+ on Windows; Unix JDK 21 unchanged).
+        org.replikativ/konserve                     {:mvn/version "0.9.393"}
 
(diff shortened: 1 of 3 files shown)


whilo:
Updated to released Konserve 0.9.393 (including the HTTP-server override) and aligned explicit Geheimnis pins to 0.2.39: the previous 0.2.33 pin caused a missing aead-encrypt-sync compile error. Merged current main into this branch. Verified released jar contains Windows binding classes and native metadata. Local directory-sync + writer-error regression suites pass: 4 tests, 23 assertions. Native workflow already uses GraalVM Community JDK 25, including Windows, but currently triggers only on main pushes; PR CI does not establish actual Windows dthk native integration. That validation remains pending. Documented the Windows JVM file-store JDK 22+ requirement.

PR #1078, opened 2026-09-07
Bound secondary-index backfill journals with local disk scratch

whilo:
## Summary

Replace the cumulative in-memory secondary backfill delta vector with JVM
disk scratch journals. Database values carry immutable accepted-prefix
descriptors rather than retaining all decoded notifications.

- Stage capture only after transaction predicates accept the report, before
  the writer chains its speculative result or enqueues it for commit.
- Enforce streaming frame-byte and per-index disk quotas, with rollback on
  partial append failure. Pure `d/with` remains I/O-free.
- Replay exactly the install operation's prefix, including when writes and
  installation queue while commits are paused.
- Retire journal files after durable publication; clean owned files on normal
  shutdown. Preserve local-exclusive ownership and rebuild-on-reconnect.
- Document configuration and operational limits in the secondary-index guide.

## Scope

This bounds the retained journal, not secondary adapter memory or transaction
input. Install still replays the full accumulated journal in the writer.
Background catch-up, a capped final replay tail, and online AVET construction
remain follow-ups. No comparator/index-format changes or resumable-build claim.
Abrupt process death can leave scratch files; fatal JVM errors may require
explicit release. Configurable byte limits reject affected transactions rather
than silently discarding notifications.

## Validation

- Focused journal, capture and lifecycle tests: 22 tests / 139 assertions passed.
- Real disk experiment: 539,623,424 bytes / 131,072 frames written and replayed
  under a 256 MiB JVM heap; scratch removed after completion.
- CLJS: 295 tests / 1,847 assertions passed.
- Whole-repository reflection check clean (111 namespaces); Java bindings compiled.
- Instrumented JVM: 1,296 tests / 14,781 assertions passed.
- Full non-spec JVM: 2,912 tests / 31,187 assertions, zero failures/errors.

The local full matrix previously hit an HTTP EOF immediately after the
permissions fixture restarted its server. Baseline full PSS and repeated
isolated candidate tests passed, so the underlying cause was not established.
The fixture now owns a fresh HTTP client and OS-assigned port per server
incarnation, preserving all persistence/authorization assertions without
adding production retries. Five namespace repetitions (180 assertions) and
the final full candidate matrix pass. This tests persistence through a new
connection, not transparent recovery of a pooled socket to a stopped server.

Tests cover retained snapshots, byte quotas, codec failure rollback, multi-index
partial staging failure, predicate rejection, pure preview, early reduction,
cleanup, and queued writes→install→writes with commit deliberately paused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ When something is added, it's typically marked *Experimental*. When the API cont
 
 ## 0.8
 
+- **Disk-backed secondary backfill journals.** JVM local-exclusive writers
+  capture concurrent changes in bounded-frame scratch files instead of retaining
+  decoded changes across database snapshots. Configure `:backfill-journal` on
+  the writer to set a scratch directory, frame limit and per-index disk quota.
+  Exceeding a limit rejects the transaction without publishing it. Installation
+  still replays the full journal; this does not bound adapter memory or enable
+  background AVET construction.
+
 - **Named transaction predicates and transaction-scoped index backfill.**
   Independent consumers can register store predicates without replacing each
   other; `ensure-tx-pred!` atomically installs an empty named slot or rejects a
--- a/doc/secondary-indices.md
+++ b/doc/secondary-indices.md
@@ -413,15 +413,44 @@ from abandoned database values.
 
 ### Backfill scalability
 
-The current beta path keeps the writer available during the snapshot scan. It
-records concurrent changes in an in-memory, per-index delta journal, then
-replays that journal in a short serialized install operation. This closes the
-lost-write race for mutable and immutable adapters, but the journal is not
-bounded. Registering an index while sustained write volume greatly exceeds
-backfill throughput can therefore consume substantial memory.
+The snapshot scan runs in the background. On the JVM, concurrent changes are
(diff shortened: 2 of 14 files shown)


whilo:
Follow-up 5e34cbce isolates the existing JVM file journal behind an opaque-prefix facade. File paths, framing, codec, truncation and quotas remain in the file backend; capture now uses a descriptor identity accessor. The logical contract specifies exact accepted-prefix replay, failure visibility, ownership and suffix abandonment without requiring physical truncation. The facade remains explicitly synchronous; no konserve backend or async/CLJS support is claimed.

Validation: independent review approved; original file corruption/quota/rollback tests preserved; opaque Object-descriptor capture and logical-prefix tests added. Focused: 13 tests / 82 assertions, green. Full non-spec JVM matrix: 2,920 tests / 31,243 assertions, zero failures. Java compilation, changed-file formatting and diff checks passed. No fresh CLJS or full instrumented-spec matrix was run for this JVM-only extraction.

Stack integration: #1079 must move its reduce-range implementation into the file backend when rebased. #1081 still has numeric byte-offset cursor assumptions; those need a journal-owned cursor interface before a non-file backend can be supported. Please do not treat the dependent PRs as retested against this new base yet.

